PR #1, opened 2019-08-14
IA-1116 terra-jupyter-base

Qi77Qi:
most scripts are straight copy from leo repo. The Dockerfile itself is the main file that changed
Screen Shot 2019-08-19 at 3 32 36 PM


build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Example: ./build.sh terra-jupyter-base
+
+IMAGE_DIR=$1
+
+TAG_NAME=$(git log --pretty=format:'%h' -n 1)
+docker image build ./$IMAGE_DIR --tag us.gcr.io/broad-dsp-gcr-public/$IMAGE_DIR:$TAG_NAME
+docker push us.gcr.io/broad-dsp-gcr-public/$IMAGE_DIR:$TAG_NAME
```
rtitle: This script must assume you're already authenticated with GCR?

Qi77Qi: yea...the script is mainly used in jenkins. So auth part is done there. I'll add a comment in the script for people who run it locally

terra-jupyter-base/Dockerfile
```diff
@@ -14,13 +14,7 @@ ENV DEBIAN_FRONTEND noninteractive
 ENV DEBIAN_REPO http://cdn-fastly.deb.debian.org
```
rtitle: I feel like the DEBIAN vars are obsolete now that we're based off Ubuntu.

terra-jupyter-base/Dockerfile
```diff
@@ -98,29 +99,10 @@ RUN useradd -m -s /bin/bash -N -u $WELDER_UID $WELDER_USER
...
- && python get-pip.py \
- && pip install ipykernel==4.10.0 \
- && python2 -m ipykernel install --name python2 --display-name "Python 2" \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
+ENV PYTHONPATH $PYTHONPATH:python:$JUPYTER_HOME/custom
 
 # Python3, kernel, and pip
-ENV PYTHON_VERSION 3.6.8
+ENV PYTHON_VERSION 3.7.0
```
rtitle: I think 3.7.4 is the latest: https://www.python.org/downloads/

terra-jupyter-base/Dockerfile
```diff
@@ -98,29 +99,10 @@ RUN useradd -m -s /bin/bash -N -u $WELDER_UID $WELDER_USER
...
 
 # lifted from https://github.com/docker-library/python/blob/dd36c08c1f94083476a8579b8bf20c4cd46c6400/3.6/stretch/Dockerfile
```
rtitle: Just wondering, should we avoid building python from source? We only did that because 3.6.8 wasn't in debian's repo. A 3.7 binary should be available, I think. 

However it doesn't hurt anything to build from source, so I could go either way.

Qi77Qi:
@rtitle think I addressed all comments...

terra-jupyter-base/Dockerfile
```diff
@@ -1,26 +1,15 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-# this is pushed to dockerhub as jmthibault79/leonardo-jupyter-min:joel-tmp-20190809-9
-
-FROM debian:stretch
-
+FROM marketplace.gcr.io/google/ubuntu1804:latest
```
rtitle: Interesting, using a Google Cloud Marketplace base image. Seems reasonable to me.

I'm slightly concerned about using the `latest` tag -- I wonder if we should tie to a particular hash to avoid unexpected upstream changes. Or is that not likely if we're staying on the same version of Ubuntu?

Qi77Qi: Bernick suggested we start using google maintained base images for security concerns...I wasn't able to find other tags for this image, since we're prebuilding these docker images, so things shouldn't all of sudden blow up in prod at least. And it's probably good to keep up-to-date base ubuntu image.

rtitle: Yeah, that makes sense 👍

PR #15, opened 2019-09-09
linting performed

jdcanas:
needs to be tested in leo


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+        
+    - name: linting
+      run: |
+        cd terra-jupyter-base/custom
+        npm install
+        node_modules/.bin/eslint --fix *.js
```
jdcanas: This .bin is how node wires packages to your project under the covers, so we can abuse this to avoid needing the permissions to add eslint to the path via `npm install -g`

terra-jupyter-base/custom/extension_entry_jupyter.js
```diff
@@ -1,7 +1,7 @@
 define([
-    'base/js/events'
+  'base/js/events'
 ], function(events) {
-    require(['custom/google_sign_in'])
-    require(['custom/edit-mode'])
-    require(['custom/safe-mode'])
+  require(['custom/google_sign_in']);
+  require(['custom/edit-mode']);
+  require(['custom/safe-mode']);
```
Qi77Qi: do we now fail build if lint fails?

jdcanas: yes

PR #16, opened 2019-09-12
[IA-1303] Jc jenkinsfile

jdcanas:
Jenkinsfile experiment complete. This may be a bit wonky since I need to actually merge this to finalize it in jenkins. 

Next steps would be to move groovy functions to a different file, but I think that would be better left as a follow-up task since this is all self-contained at the moment.

I have tested all pieces end to end, but I was wary of building images without legitimate changes, so I'm going to use my linting PR as a final test after this. 

Things I learned from Jenkinsfile experiment:
	
- allows for much cleaner syntax/maintenance than walls of bash for complicated build tasks. I much prefer my `replaceTags` function to https://github.com/broadinstitute/dsp-jenkins/blob/master/jobs/fc-jenkins/LeonardoPublishDataprocImage.groovy#L62
- much cleaner shell syntax 
- powerful conditional flow abilities
- great flexibility over build display (the calling of `setBuildDescription` and `setResult` provide feedback directly in the jenkins UI as opposed to having to dig through build logs, see https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-build-terra-docker/ for fancy colors and messages )
- master/slave node implementation details can be exposed when you get into complex stuff (see FilePath)
    - some code gets run on master node, which can be very confusing to debug
- traditional functional iterators (i.e. `foreach`, `filter`) present problems due to Jenkins features regarding serialization along the build process for reproducibility (see for loops)  
    - can be worked around with `@nonCPS` annotation, but this brings more problems than seemed worth it
- permissions issues with a lot of methods made me need to whitelist methods here https://fc-jenkins.dsp-techops.broadinstitute.org/scriptApproval/
     - Could be a security vulnerability, I'd like @RLuckom to weigh in, but I don't think so since our jenkins instance is private, so if you're running things there or modifying jobs, you've already got access. More annoying for developers it seems.
- other than the above, all token access and the like seems to be as secure as DSL jobs.


jenkins/Jenkinsfile
```diff
@@ -0,0 +1,240 @@
...
+
+    println("printing changes in $currConfig and running git add")
+    sh(
+      script: "git diff $currConfig; git add $currConfig"
+    )
+  }
+}
+
+//parameterize with ":" prefix and end quote suffix instead of hardcode?
+def void replaceTags(String filePath, LinkedHashMap<String, String> replacementMap) {
```
Qi77Qi: can you add comment that has an example of what the tag looks like before and after this function?

jenkins/Jenkinsfile
```diff
@@ -0,0 +1,240 @@
...
+
+    stage('Docker Auth') {
+      steps {
+        //the chown is necessary because something on the jenkins instance keeps touching the logs with sudo and changing the owner to root
+        //this which causes gcloud auth command to fail
+        sh """
+          sudo chown -R jenkins:jenkins /home/jenkins/.config
+          docker run --rm  -v /etc/vault-token-dsde:/root/.vault-token:ro broadinstitute/dsde-toolbox:latest vault read --format=json secret/dsde/dsp-techops/common/dspci-wb-gcr-service-account.json | jq .data > dspci-wb-gcr-service-account.json   
+          gcloud auth activate-service-account --key-file=dspci-wb-gcr-service-account.json
+          gcloud auth configure-docker --quiet
```
Qi77Qi: we probably need to fix this job to the same node we run welder publish image cuz `gcloud auth configure-docker` causes trouble for other jenkins jobs

jenkins/Jenkinsfile
```diff
@@ -0,0 +1,240 @@
...
+
+//used to exit early if no changes are detected
+def shouldAbortBuild = false
+
+//declared in outer scope so it persists between steps
+//omitting the 'def' is what makes it global
+LinkedHashMap<String, String> versionMap = [:]
+
+def ArrayList<String> getChangedImages() {
+  def String[] images = readFile("jenkins/imageDirs.txt").trim().split("\n")
```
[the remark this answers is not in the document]
jdcanas: I like this to be something explicit personally, I think it's a better build practice to have the logic/detection be as minimal as possible, and tell it what to do through configs/files in the repo. It would already be pretty messy since we'd have to exclude a 3 different dirs, and if anyone adds any dirs that aren't images it would cause the job to fail.

I'd rather the failure case be it doesn't build an image because we didn't update a config rather than the job randomly starts failing because of an implementation detailed squirreled away in the job logic.

Also, it allows us flexibility in the sense that, if for some reason, we want to lock a version for a while or only detect changes on a specific image, we can just update the config.

jenkins/Jenkinsfile
jdcanas: the 'agent' line at the top of the file does this. I'll add a comment

jenkins/Jenkinsfile
```diff
@@ -0,0 +1,244 @@
...
+   agent { label 'node210' }
+
+  options { 
+    //we don't want to be building more than one PR's changes at once, we will queue them instead
+    disableConcurrentBuilds() 
+  }
+
+  triggers {
+       // poll repo every 2 minute for changes
+       pollSCM('*/2 * * * *')
```
Qi77Qi: shouldn't it be triggerred on PR merge to master branch instead of polling?

PR #18, opened 2019-09-18
Update nodejs.yml

jdcanas:
using new git actions feature. this file is more or less auto-generated.

Note these changes are untested as apparently it needs to be in the repo to take effect, so I limited the scope to a test branch for now. Therefore, there will need to be a second PR for the final version.


jdcanas:
made a legit PR

PR #19, opened 2019-09-23
Make Bioconductor version more explicit for developers to see

nturaga:
Based on a conversation with Justin and Qi, this change is being implemented. Tested on Leo locally.

cc @Qi77Qi


Qi77Qi:
thanks @nturaga

PR #21, opened 2019-09-25
[IA-1235] Jupyter-R-GATK-Samtools Image

akarukappadath:
Terra image that contains everything from terra-python and terra-R images with GATK and Samtools. It has been verified that Samtools version supports GCS.


terra-jupyter-gatk/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+## 0.0.1 - 09/24/2019
+
+- added all from terra-jupyter-python:0.0.1
```
Qi77Qi: can you add a readme file similar to https://github.com/DataBiosphere/terra-docker/tree/master/terra-jupyter-base

PR #43, opened 2019-10-08
[IA-1428] Jc stagger meta lock

jdcanas:


terra-jupyter-base/custom/edit-mode.js
```diff
@@ -106,21 +106,21 @@ define(() => {
...
+            checkMeta(true);
         }, lastLockedTimer);
 
         window.onbeforeunload(function() {
             clearInterval(syncMaintainer);
         });
     }
 
-    function checkMeta() {
+    function checkMeta(shouldStaggerCalls = true) {
```
jdcanas: needed because we want the first lock call to occur immediately after check meta, and subsequent calls to be staggered

terra-jupyter-base/custom/edit-mode.js
```diff
@@ -186,14 +186,16 @@ define(() => {
...
+    function handleMetaSuccess(res, shouldStaggerCalls) {
         toggleMetaFailureBanner(false); // sets banner for meta status
 
         const isEditMode = res.syncMode == 'EDIT';
 
         if (isEditMode) {
             handleCheckMetaResp(res); // displays modal if theres an issue in the payload
-            getLock(res); // gets lock if in edit mode
+            //if its not the first time we are acquiring lock, wait half the time until next check meta call before calling this.
+            waitTime = shouldStaggerCalls ? lastLockedTimer / 2 : 0
```
jdcanas: Currently, checkMeta occurs every 60 seconds and autosave occurs every 5 seconds. the /2.2 ensures we stagger by approximately 27 seconds, which should be off-cycle to both.

terra-jupyter-base/custom/edit-mode.js
```diff
@@ -186,14 +186,17 @@ define(() => {
...
-    function handleMetaSuccess(res) {
+    function handleMetaSuccess(res, shouldStaggerCalls) {
         toggleMetaFailureBanner(false); // sets banner for meta status
 
         const isEditMode = res.syncMode == 'EDIT';
 
         if (isEditMode) {
             handleCheckMetaResp(res); // displays modal if theres an issue in the payload
-            getLock(res); // gets lock if in edit mode
+            //if its not the first time we are acquiring lock, wait half the time until next check meta call before calling this.
```
calbach: This might work , but I'm having difficulty following the approach. I think it would be simpler to do the following at L109:

```
// Offset from autosave timer to mitigate Welder concurrency issues.
const initialSyncDelay = 27000;

...
checkMeta();
setTimeout(() => {
  checkMeta();
  initSyncMaintainer();
}, initialSyncDelay);
```

jdcanas: The approach I took staggers the lock from checkMeta and autosave. The approach you mention staggers checkMeta + lock from autosave. I've changed it to the approach you suggest, since it seems that's what we want to be doing.

Qi77Qi:
@calbach , btw, this update is only applicable to new jupyter images. So you guys will need to use one of the base images in cluster creation request in order to pick up the change

terra-jupyter-base/custom/edit-mode.js
```diff
@@ -107,7 +107,10 @@ define(() => {
         }
 
         checkMeta();
-        initSyncMaintainer();
+
+        const initialSyncDelay = 27000;
+        //we want subsequent checkMeta + lock calls to be off-cycle of autosave
+        setTimeout(() => initSyncMaintainer(), initialSyncDelay);
```
calbach: Note: checkMeta will trigger at the following times with this implementation: 0s, 1m27s, 2m27s, ... 

I think this is probably fine (vs also having a check @ 27s), but just calling it out to make sure we're on the same page.

Qi77Qi:
can you update version for all the other images as well? or at least terra-jupyter-python, assuming that's the image AOU will use

jdcanas:
> or at least terra-jupyter-python, assuming that's the image AOU will use

Sure. This will cause a version cascade though. Updated the appropriate files.

terra-jupyter-base/custom/edit-mode.js
jdcanas: Yeah, I feel that its fine to wait a bit after the first call, as such a small difference shouldn't affect the integrity and it makes the code cleaner.

PR #48, opened 2019-10-17
[IA-1417] Jc update jenkinsfile

jdcanas:
- Update build script to be able to correctly source the vault token regardless of if it is run on jenkins or local (this is redundant, as the jenkins job gets the token, but I'd rather have redundant defensive code than auth errors)
- Fix issue where the job does not properly order image builds
- Job will now fail appropriately when an image fails to build
- Disable doc generation for gatk until pip is working
- Job will now create PRs
- Dockerimage for hub, the tool used to create PRs via CLI, is added

### DIFF
--- a/hub/Dockerfile
+++ b/hub/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:14.04
+
+RUN apt-get -y update 
+
+RUN apt-get -y install software-properties-common
+
+RUN add-apt-repository -y ppa:cpick/hub
+
+RUN apt-get -y update
+
+RUN apt-get -y install hub
+
+RUN git config --system core.logallrefupdates false
+
+RUN mkdir -p /app/leo
+WORKDIR /app/leo
+
+RUN useradd -m dsde-jenkins
+USER dsde-jenkins
+
+RUN mkdir /home/dsde-jenkins/.config
+
+CMD ["/bin/bash"]
+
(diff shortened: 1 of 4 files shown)


hub/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:14.04
```
Qi77Qi: what is this for?

Qi77Qi:
can you add some brief description as to what this PR does

jenkins/imageDirs.txt
```diff
@@ -1,6 +1,5 @@
 terra-jupyter-base
-terra-jupyter-bioconductor
-terra-jupyter-gatk
-terra-jupyter-hail
 terra-jupyter-python
 terra-jupyter-r
+terra-jupyter-bioconductor
+terra-jupyter-gatk
```
Qi77Qi: do you still want to build gatk now that it's not really working?

hub/Dockerfile
jdcanas: this is the dockerfile used to make PRs

jenkins/imageDirs.txt
jdcanas: I'll keep the build since the image builds, but I do not generate documentation. I think its worth keeping since the image seems to work fine, pip is just not in the path correctly (which is what our documentation depends on)

PR #49, opened 2019-10-18
IA-1406 fix google sign in discrepancy

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1406


terra-jupyter-python/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.0.3 - 10/18/2019
+
+- Update `terra-jupyter-base` version to `0.0.4`
+`Image URL: us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:0.0.4`
```
akarukappadath: this should be `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.3` right?

terra-jupyter-base/scripts/kernel/kernel_bootstrap.sh
```diff
@@ -6,13 +6,13 @@
...
-#   ~jupyter-user/<workspace-name>/notebook.ipynb
+#   ~jupyter-user/<workspace-name>/<mode>/notebook.ipynb
 #
 # It exploits the fact that the CWD of a launching notebook is named after the workspace.
 # If notebooks are ever launched from other directories, this script will break.
 
 # The workspace name is simply the CWD of the running notebook.
 PWD="$(pwd)"
-export WORKSPACE_NAME="$(basename "$PWD")"
+export WORKSPACE_NAME=`basename $(dirname $(pwd))`
```
Qi77Qi: this fixes the `WORKSPACE_NAME` issue after welder rollout

Qi77Qi:
@rtitle , we'd have to update it manually. https://broadworkbench.atlassian.net/browse/IA-1446 will fix the problem

melissachang:
Hi, I'm running into WORKSPACE_NAME bug on app.terra.bio. When will this fix be on app.terra.bio?

rtitle:
Hi @melissachang ,

We're targeting the fix for Monday's release (10/28). Once released you'll need to stop/start your runtime to get the fix.

In the meantime, this is a workaround to get the workspace name in a python variable:

```
import os
workspace = os.path.basename(os.path.dirname(os.getcwd()))
```

terra-jupyter-base/scripts/kernel/kernel_bootstrap.sh
mbookman: Could the comment in the code here be updated? It is no longer accurate:

```
# The workspace name is simply the CWD of the running notebook.
```

Should be something like:

```
# Notebooks are copied into a subdirectory, such as "edit" or "safe", which is the CWD.
# The parent directory is the name of the workspace.
```

Thanks!

PR #50, opened 2019-10-23
[IA-1467] Fix WORKSPACE_NAME ev for workspace names with spaces

rtitle:
https://github.com/DataBiosphere/leonardo/pull/1114
https://broadworkbench.atlassian.net/browse/IA-1467


terra-jupyter-base/scripts/kernel/kernel_bootstrap.sh
```diff
@@ -13,6 +13,6 @@
 
 # The workspace name is simply the CWD of the running notebook.
 PWD="$(pwd)"
-export WORKSPACE_NAME=`basename $(dirname $(pwd))`
```
Qi77Qi: I think the second quote is off...it should be after `)`

rtitle: Right you are, fixed

PR #51, opened 2019-10-29
[IA-1446] Jc build script upgrade

jdcanas:
Upgrades the build scripts to produce and upload json documentation for the UI.

In addition, a json file will be produced and uploaded which states the current versions, with all the info the UI needs to display and links to the individual image documentation.

This is accomplished via 2 python scripts (one for each purpose) and a config file. 

This required small changes in both the terra-docker image building jenkins job, and the custom dataproc job as a result of more flexibility with the conf.


Qi77Qi:
cc @zarsky-broad about this PR in case he has comments

jdcanas:
Need some additional testing jenkins-side before finalizing this. I will make a PR to terra-ui with the changes necessary. I'll reach out to you directly with some context when I do @zarsky-broad, but this is not a breaking change to what terra-ui currently has.

zarsky-broad:
Does this generate separate files of different types of dependencies? I.e. for the default image, there only appears to be one list of packages, not broken up by type (also, `terra-docker-image-documentation/leonardo-jupyter-dev-versions.json` as currently in the bucket isn't valid JSON, there's a syntax error with misplaced quotes)

jdcanas:
> Does this generate separate files of different types of dependencies? I.e. for the default image, there only appears to be one list of packages, not broken up by type (also, `terra-docker-image-documentation/leonardo-jupyter-dev-versions.json` as currently in the bucket isn't valid JSON, there's a syntax error with misplaced quotes)

No, I condense them into one file but hadn't gotten to separating them by type at the time you commented. Good catch on the .json, I had to manually create that one by hand and some IDE tricks. I'm surprised my IDE didn't catch it, I must have fat fingered something after I linted.

scripts/generate_package_docs.py
```diff
@@ -0,0 +1,98 @@
...
+
+  return json_python_doc
+
+def get_gatk_doc(params):
+  cmd = "ls /etc/ | grep gatk | tr '-' ' ' | awk '{print $2}'"
+  gatk_version = utils.docker_exec(params.image_dir, cmd)
+
+  return [{"gatk": gatk_version}]
+
+class ParamProcessor():
```
kyuksel: Instead of rolling own, it may be better to use a library. I used [ArgumentParser](http://zetcode.com/python/argparse/) on another library and it worked fairly well. It's the most helpful when centrally listing the args in a separate file. I understand that these are scripts so I'll let you judge the benefit of it  😊

scripts/generate_package_docs.py
```diff
@@ -0,0 +1,98 @@
...
+import re
+import utils
+
+config_location = "config/conf.json"
+
+def main():
+  params = ParamProcessor(argv)
+  write_docs_to_bucket(params)
+
+def write_docs_to_bucket(params):
```
kyuksel: I like being able to see what parameters are used by a method by looking at the method's signature, which is difficult to do when all `params` are passed in. It'd also allow for including types in method signatures...which is a good segue to plug [type hints](https://docs.python.org/3/library/typing.html) and [mypy](http://mypy-lang.org/). Again, because these may be stand-alone scripts, they may not be a big deal but...

scripts/generate_package_docs.py
jdcanas: ooooh I didn't know about this built-in module this is great. I've always stayed away from complex flags, but now I'm itching for a use case. I think I'm going to leave it in this particular case, since it's fairly basic and there's no optional params, plus a bit of custom processing is needed with the params. Good callout though, I'll definitely keep this in mind over the custom ones I've been rolling.

scripts/generate_package_docs.py
jdcanas: i implemented all the typing hints only to learn that this became a thing in python 3.5 so I can't since I had to revert to 2.7 for jenkins :(

kyuksel: Bummer :( Could still spell out the needed subset of arguments but not a blocker.

PR #54, opened 2019-11-06
Avoid r-recommended packages being installed in /usr/lib/R/library 

Qi77Qi:
1. r-base: this installs the r-recommended packages since it's a
dependency and these packages get installed in "/usr/lib/R/library".

2. r-recommended: this basically does the same thing as installing the
packages in the /usr/lib/R/library location.

We want to avoid installing the r-recommended packages, so both r-base
and r-recommended are removed. To allow installation to happen for these
recommended packages in /home/jupyter-user/.rpackages, they are
individually listed in the BiocManager::install command.

The "ask" option is added as FALSE to the BiocManager, as we don't want
to update unprompted. See below,

 ask: 'logical(1)' indicating whether to prompt user before
      installed packages are updated.  If TRUE, user can choose
      whether to update all outdated packages without further
      prompting, to pick packages to update, or to cancel updating
      (in a non-interactive session, no packages will be updated
      unless 'ask = FALSE').
This solves Bioconductor#2

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -17,7 +17,7 @@
             "base_label": "Bioconductor",
             "tools": ["python", "r"],
             "packages": { "r": ["BiocVersion", "tidyverse"] },
-            "version": "0.0.4",
+            "version": "0.0.5",
             "automated_flags": {
                 "generate_docs": true,
                 "include_in_ui": true,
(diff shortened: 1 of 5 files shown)


nturaga:
Closing this PR in favor of #52

PR #59, opened 2019-11-21
changing UI label

jdcanas:
Changes the static portion of the label in the UI for the base image from 'minimal' to 'base'

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -53,7 +53,7 @@
         },
         {
             "name": "terra-jupyter-base",
-            "base_label": "Minimal",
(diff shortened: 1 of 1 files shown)


Qi77Qi:
do we still need the change now that we don't want to display in UI?

jdcanas:
> do we still need the change now that we don't want to display in UI?

I say we still should. The fields are independent and I'd totally forget it we want to include it at some point if extending it becomes popular. TBH, we could feasibly advertise it as the fastest if you don't need a bunch of junk and are starting off exploring.

PR #64, opened 2019-11-27
include gatk image in custom dataproc image

Qi77Qi:
One theory about why AOU cluster startup increased since 11/19 is that they started to use terra-jupyter-gatk image, and we saw huge increased startup time since then in perf (altho it doesn't seem to affected prod much)

Screen Shot 2019-11-27 at 11 58 02 AM

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -84,10 +84,10 @@
             "packages": {},
             "version": "0.0.6",
             "automated_flags": {
-                "include_in_ui": true,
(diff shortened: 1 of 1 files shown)


config/conf.json
```diff
@@ -87,7 +87,7 @@
                 "include_in_ui": true,
                 "generate_docs": false,
```
jdcanas: if we want this included in the UI, the docs need to be generated

PR #65, opened 2019-11-27
Misc bug fixes/updates

jdcanas:
- Updates README with terra-ui interaction details
- Doc generation script now references remote documentation instead of local
- leonardo-build-terra-docker will now archive a file that reports which images are built to determine how far it got


jenkins/Jenkinsfile
```diff
@@ -354,6 +357,12 @@ pipeline {
          "\nIf you would like to check the status of the PR, click: $prLink")
       }
     }
+
+    post {
+      always {
+          archiveArtifacts artifacts: "$helpfulOutputFile"
```
Qi77Qi: 👍

PR #71, opened 2019-12-12
Install r script

nturaga:


Qi77Qi:
update https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L20 :)

nturaga:
> update https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L20 :)

Knew I forgot something :D

PR #76, opened 2019-12-19
ignore failure if it fails to remove a file

Qi77Qi:


Qi77Qi:
cc @jdcanas in case this isn't ideal

PR #85, opened 2020-01-11
IA-1608 upgrade gatk version

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1608

Tested with https://github.com/DataBiosphere/leonardo/pull/1209

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -82,7 +82,7 @@
             "base_label": "GATK",
             "tools": ["python", "gatk"],
             "packages": {},
-            "version": "0.0.7",
+            "version": "0.0.8",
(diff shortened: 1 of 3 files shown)


jdcanas:
> 👍
> 
> This should trigger a job to update https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/resources/reference.conf#L4, right?

It will, but gatk will not be changed/updated unless it appears in the format the job is looking for, aka the string that is used for the directory

Qi77Qi:
@rtitle I think it should...will keep an eye on the a followup PR

PR #88, opened 2020-01-13
IA-1610 Include hail and bioc in the dataproc image

rtitle:
Friends with https://github.com/DataBiosphere/leonardo/pull/1214


Qi77Qi:
updated title with ticket number so that all PRs for that ticket is shown in jira

PR #90, opened 2020-01-14
IA-1555 IA-1616 use just base label

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-1555

@sharmatime suggested that we want just the base label


Qi77Qi:
looks like we're not doing this anymore...

PR #93, opened 2020-01-15
Add pandas_profiling (Python) and reticulate (R) to GATK image for AoU

jmthibault79:


terra-jupyter-r/Dockerfile
```diff
@@ -37,6 +37,11 @@ RUN mkdir -p /home/jupyter-user/.rpackages \
     "remotes", \
     "BiocManager", \
     "devtools"))' \
+
+    # fix for AoU-blocking bug https://github.com/rstudio/reticulate/issues/693
+    # after version 1.15 is released to CRAN, we can install reticulate in the standard way
+    && R -e 'devtools::install_github("rstudio/reticulate", ref="00172079")' \
```
Qi77Qi: I don't know a lot about R packages. @nturaga does this look right to you or should we be using `BiocManager`?

Qi77Qi:
thanks @jmthibault79 ...Can you also publish a test gatk image manually and send a pr similar to https://github.com/DataBiosphere/leonardo/pull/1209 to trigger automation tests...

terra-jupyter-r/Dockerfile
jmthibault79: This is a bit of a hack because I need to install a specific git hash: the fix is not yet released.

I'm not familiar w/ bioc so I don't know if there's an equivalent command.

jmthibault79:
@Qi77Qi https://github.com/DataBiosphere/leonardo/pull/1222

terra-jupyter-r/Dockerfile
nturaga: You can use `BiocManager`. 

It's just `BiocManager::install("rstudio/reticulate", ref="00172079")`

jmthibault79: Thanks!

I'm not familiar with the implications regarding plain R vs. installing via Bioc.  Should I update this as suggested?

Qi77Qi: I think all the other packages are installed via `BiocManager`. Might as well use the same?

jmthibault79: ok

jmthibault79:
Updated this to use BiocManager and pushed the new image to https://github.com/DataBiosphere/leonardo/pull/1222

PR #97, opened 2020-01-17
revert label change

Qi77Qi:
the label seems being used in the script for referencing...reverting the change for now


Qi77Qi:
I looked a bit and then give up...created https://broadworkbench.atlassian.net/browse/IA-1624 for now

PR #99, opened 2020-01-28
[IA-1637] Hail jobs fail with "mismatched python version"

rtitle:
The main fix is to specify the `PYSPARK_PYTHON=python3` env variable in the kernel definition. Otherwise it seems to default to python 2.7. See: https://cloud.google.com/dataproc/docs/tutorials/python-configuration

### DIFF
--- a/terra-jupyter-hail/Dockerfile
+++ b/terra-jupyter-hail/Dockerfile
@@ -5,7 +5,7 @@ COPY scripts $JUPYTER_HOME/scripts
 
 ENV PIP_USER=false
 ENV PYTHONPATH $PYTHONPATH:/usr/lib/spark/python
-ENV HAIL_VERSION=0.2.31
+ENV HAIL_VERSION=0.2.30
 
 RUN find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
     && $JUPYTER_HOME/scripts/kernel/kernelspec.sh $JUPYTER_HOME/scripts/kernel /usr/local/share/jupyter/kernels \
--- a/terra-jupyter-hail/Dockerfile
+++ b/terra-jupyter-hail/Dockerfile
@@ -5,7 +5,7 @@ COPY scripts $JUPYTER_HOME/scripts
 
 ENV PIP_USER=false
 ENV PYTHONPATH $PYTHONPATH:/usr/lib/spark/python
-ENV HAIL_VERSION=0.2.30
+ENV HAIL_VERSION=0.2.31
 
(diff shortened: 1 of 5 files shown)


terra-jupyter-hail/Dockerfile
```diff
@@ -24,7 +28,9 @@ RUN mkdir -p /etc/spark/conf.dist && mkdir -p /etc/hadoop/conf.empty && mkdir -p
     && (cd $X && pip3 download hail==$HAIL_VERSION --no-dependencies && \
         unzip hail*.whl &&  \
         grep 'Requires-Dist: ' hail*dist-info/METADATA | sed 's/Requires-Dist: //' | sed 's/ (//' | sed 's/)//' | grep -v 'pyspark' | xargs pip install) \
-    && rm -rf $X
+    && rm -rf $X \
+    && apt-get clean \
```
rtitle: Added this, removes ~2G from the image

terra-jupyter-hail/Dockerfile
```diff
@@ -1,13 +1,17 @@
 FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.8
 USER root
-ENV PIP_USER=false
 
+COPY scripts $JUPYTER_HOME/scripts
```
rtitle: This is a similar pattern to the [terra-jupyter-r](https://github.com/DataBiosphere/terra-docker/tree/master/terra-jupyter-r) Dockerfile, which also adds a custom kernepsec.

terra-jupyter-hail/Dockerfile
kyuksel: nice!

terra-jupyter-hail/scripts/kernel/python_kernelspec.tmpl
```diff
@@ -0,0 +1,15 @@
+{
+ "language": "python",
+ "display_name": "Python ${PY_VERSION}",
```
Qi77Qi: just curious..where is `PY_VERSION` set?

rtitle: `kernelspec.sh` replaces it with "3": https://github.com/DataBiosphere/terra-docker/pull/99/files#diff-d7fa6e149ccb1fc5c30017df315a618cR7

I sort of want to remove `kernelspec.sh` -- now that we have custom dockers (and fewer kernels) we could just `ADD` the `kernel.json`s in the right places directly in the Dockerfile. But changing that pattern would touch the base image + init-actions.sh so I felt it's a bit out of scope of this PR.

PR #100, opened 2020-01-28
Update Hail to 0.2.31

tpoterba:


Qi77Qi:
just to double check, this upgrade doesn't require spark upgrade?

2 other things you need to update
1. Bump image version in https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L33
2. Update https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-hail/CHANGELOG.md

There's an open PR on hail image https://github.com/DataBiosphere/terra-docker/pull/99, maybe need to coordinate a bit?

cseed:
@Qi77Qi We should also include the Hail changelog, so Terra users can see what's in the new Hail version.  Can you point us to where that goes?  I'm maybe remembering it goes in a different repo?  Our changelog can be seen deployed in the Hail docs here: https://hail.is/docs/0.2/change_log.html#version-0-2-31.

Qi77Qi:
I think putting hail's changelog link in https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-hail/CHANGELOG.md should be fine

rtitle:
Oops just saw this - @tpoterba I was also going to merge a change to the Hail image in https://github.com/DataBiosphere/terra-docker/pull/99, I could include this if you want

tpoterba:
er, don't do this yet...discovered a possibly significant performance regression in 0.2.31. Still investigating.

PR #101, opened 2020-01-30
Jc fix build

jdcanas:
- prevents IMAGE_EXISTS being triggered on VERSIONS containing a substring of the VERSION we care about
- changes from pipe to or to avoid termination on failed commands, which was messing up line 15

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,7 +10,7 @@ TAG_NAME=$(git log --pretty=format:'%h' -n 1)
 GCR_IMAGE_REPO=$(cat config/conf.json | jq -r .gcr_image_repo)
 
 #for some reason, this command fails if the script is in strict mode because grep not finding something exits with 1
-#we grep for "$VERSION," because the result we are interested will always be of the form "$VERSION,$GIT_HASH" and this ensures we don't match on substrings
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -10,7 +10,9 @@ TAG_NAME=$(git log --pretty=format:'%h' -n 1)
 GCR_IMAGE_REPO=$(cat config/conf.json | jq -r .gcr_image_repo)
 
 #for some reason, this command fails if the script is in strict mode because grep not finding something exits with 1
-IMAGE_EXISTS=$(gcloud container images list-tags $GCR_IMAGE_REPO/$IMAGE_DIR | grep $VERSION) | true
+#we grep for "$VERSION," because the result we are interested will always be of the form " $VERSION,$GIT_HASH " and this ensures we don't match on substrings
+#This is subject to being broken if we tag/push the images in a different order.  
+IMAGE_EXISTS=$(gcloud container images list-tags $GCR_IMAGE_REPO/$IMAGE_DIR | grep " $VERSION,") || true
```
rtitle: This seems a little brittle.. how about something like this?
```
gcloud container images list-tags --filter="tags:0.0.7" us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python
```

jdcanas: Yeah I hated how brittle it seemed, I couldn't think of a better way at the time though. That filter option supports regex though!

PR #102, opened 2020-02-05
[IA-1653] Fix cookie setting in google_sign_in.js in terra-jupyter-base

rtitle:
Also update all sub-images with the new base.

See https://broadworkbench.atlassian.net/browse/IA-1653

### DIFF
--- a/terra-jupyter-base/custom/google_sign_in.js
+++ b/terra-jupyter-base/custom/google_sign_in.js
@@ -81,7 +81,8 @@ function startTimer() {
   setInterval(statusCheck, 60000);
 }
 
-// Note: this should match https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/CookieHelper.scala
+// Note: this should match
+// https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/CookieHelper.scala
 function setCookie(token, expiresIn) {
   document.cookie = 'LeoToken=' + token + '; Max-Age=' + expiresIn + '; Path=/; Secure; SameSite=None'
 }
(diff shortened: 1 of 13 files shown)


rtitle:
I still need to actually test this

terra-jupyter-base/custom/google_sign_in.js
```diff
@@ -81,10 +81,10 @@ function startTimer() {
   setInterval(statusCheck, 60000);
 }
 
+// Note: this should match
+// https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/CookieHelper.scala
 function setCookie(token, expiresIn) {
-  const expiresDate = new Date();
-  expiresDate.setSeconds(expiresDate.getSeconds() + expiresIn);
-  document.cookie = 'LeoToken=' + token + ';secure;expires=' + expiresDate.toUTCString() + ';path=/';
+  document.cookie = 'LeoToken=' + token + '; Max-Age=' + expiresIn + '; Path=/; Secure; SameSite=None'
```
rtitle: This is the actual fix

rtitle:
So I built the image and deployed it through Terra and confirmed that token refresh works. Though it was never actually broken for me so I'm not sure how great a test that is.

@calbach would you be able to test the image `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:ia-1653` through AoU and see if the issue is resolved? Once this PR is merged, the image version will become `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:0.0.11`.

calbach:
Thanks @rtitle , I'm lazy so I decided to just use chrome local overrides to paste in your updated JS on a Workbench test window rather than pulling the new image. I've verified that it is successfully updating the Max-Age parameter of the cookie which seems to indicate it is working - I will check back in >1h to verify it's not 401ing.

rtitle:
Great, thanks!

calbach:
Confirmed it works for at least 1.5h :+1: , thanks.

rtitle:
Great, I also left a notebook open for hours and it seems to work. Merging.

PR #103, opened 2020-02-24
Jc notebook config fallback

jdcanas:
Should be ready... I'm not sure what the google cloud build check that is failing is actually checking though...

This PR does the following:

Updates `terra-jupyter-base` and cascades the new version to all images.

In **edit-mode.js**:
- lints
- renames the saveAs function in edit-mode.js, as it seemed to be conflicting with https://github.com/eligrey/FileSaver.js, which pollutes the global namespace. Not sure if this was actually causing issues, but it was preventing me from testing our saveAs in the console. 
- moves the `require` calls at the top of the file to the function that needs it. This should prevent them from being required in views where they cannot be loaded (they were causing console errors in terminal view). This is a slightly questionable approach, as it means they are re-loaded with each function call, but the majority of the time the functions in which they are called triggers navigation to another page (and a full re-load), and the ideal approach involves parameterizing  this [script](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/scripts/extension/jupyter_install_notebook_extension.sh#L14) to take a section ([documention](https://jupyter-notebook.readthedocs.io/en/stable/extending/frontend_extensions.html)), on which there is no documention on the valid `section` options beyond `common`.

In **google_sign_in.js**:
- add conditional to `updateParams` to use the url as a fallback to get the `googleProject` and `clusterName`
- add fallback call to get `notebook.json` manually from jupyter's config api
- put xhr requests in a re-usable function


terra-jupyter-base/custom/edit-mode.js
```diff
@@ -1,443 +1,447 @@
-const dialog = require('base/js/dialog');
```
jdcanas: moving these inside the functions that use them is the only change of 2

terra-jupyter-base/custom/edit-mode.js
```diff
@@ -1,443 +1,447 @@
...
+          // then we rename the file, as POST does not allow us to specify the file name
+          fetch(params.jupyterServerApi + res.path, patchPayload)
+              .then((res) => {
+                // navigate to new file
+                window.location.href = params.jupyterFsHref + safeModeDir + '/' + originalNotebookName;
+              });
+        });
+  }
+
+  function notebookSaveAs() {
```
jdcanas: renaming this function is change 2

jdcanas:
I don't think the check is anything to worry about, I think its just concerned that the images referenced don't exist yet, and our CI should detect changes in all and build them in the correct order

terra-jupyter-base/custom/google_sign_in.js
```diff
@@ -16,16 +16,56 @@ const params = {
...
+                         Found googleProject: '${googleProject}' and clusterName '${clusterName}'`);
+
+  params.googleProject = googleProject;
+  params.clusterName = clusterName;
+
+  fallbackReadNotebookConfig(googleProject, clusterName);
+}
+
+function fallbackReadNotebookConfig(googleProject, clusterName) {
+  const url = `/notebooks/${googleProject}/${clusterName}/api/config/notebook`;
```
rtitle: Nitpick: can we use this URL format?
```
/proxy/googleProject/clusterName/jupyter/api/config/notebook
```
The /notebooks one is technically deprecated, I don't think any clients are using it anymore.

terra-jupyter-base/custom/google_sign_in.js
```diff
@@ -72,19 +127,19 @@ function startTimer() {
     setInterval(doAuth, 180000);
   });
 
-
   function statusCheck() {
-    const xhttp = new XMLHttpRequest();
-    xhttp.open("GET", "/notebooks/" + params.googleProject + "/" + params.clusterName + "/api/status", true);
-    xhttp.send();
+    const url = `/notebooks/${params.googleProject}/${params.clusterName}/api/status`;
```
rtitle: Same as above to use `/proxy/project/name/jupyter/api/status`

rtitle:
Overall looks good, a couple nits in the JS code.

Image bumps seem correct, the last few times I've done this I've had success with the CI.

terra-jupyter-base/custom/google_sign_in.js
```diff
@@ -16,16 +16,56 @@ const params = {
...
+                         Found googleProject: '${googleProject}' and clusterName '${clusterName}'`);
+
+  params.googleProject = googleProject;
+  params.clusterName = clusterName;
+
+  fallbackReadNotebookConfig(googleProject, clusterName);
+}
+
+function fallbackReadNotebookConfig(googleProject, clusterName) {
+  const url = `/proxy/${googleProject}/${clusterName}/api/config/notebook`;
```
rtitle: Need to insert `jupyter` when using the `proxy` prefix:
```
/proxy/${googleProject}/${clusterName}/jupyter/api/config/notebook
```

terra-jupyter-base/custom/google_sign_in.js
```diff
@@ -72,19 +127,19 @@ function startTimer() {
     setInterval(doAuth, 180000);
   });
 
-
   function statusCheck() {
-    const xhttp = new XMLHttpRequest();
-    xhttp.open("GET", "/notebooks/" + params.googleProject + "/" + params.clusterName + "/api/status", true);
-    xhttp.send();
+    const url = `/proxy/${params.googleProject}/${params.clusterName}/api/status`;
```
rtitle:
```
/proxy/${params.googleProject}/${params.clusterName}/jupyter/api/status
```

terra-jupyter-base/custom/package-lock.json
```diff
@@ -0,0 +1,1021 @@
+{
```
rtitle: Where does this come from?

jdcanas: its generated when I did `npm install` (so intelliJ would lint for me as I went) to ensure reproducible runs of the command (i.e. we get the same version of eslint). It's generally good to commit it, but I think I'll add it to .gitignore in this case since this isn't a real web app, and we don't want it put on clusters.

PR #104, opened 2020-02-26
[IA-1706] Install JDK 8 in the gatk image

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-1706

We currently install Java 11. Apparently GATK requires exactly Java 8:

https://gatk.broadinstitute.org/hc/en-us/articles/360035889531

> Java 8 / JRE or SDK 1.8
> The GATK is a Java-based program, so you'll need to have Java installed on your machine. The Java runtime version should be at 1.8 exactly. To be clear: we do not yet support 1.9, and older versions (1.6 and 1.7) no longer work. You can check what version you have by typing java -version at the command line. This article has some more details about what to do if you don't have the right version. Both the Sun/Oracle Java JDK and OpenJDK versions are fully supported.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -82,7 +82,7 @@
             "base_label": "New Default (released on January 14)",
             "tools": ["gatk", "python", "r"],
             "packages": {},
-            "version": "0.0.12",
+            "version": "0.0.13",
             "automated_flags": {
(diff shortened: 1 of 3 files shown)


rtitle:
I manually built this and it seems to work. Will merge.

PR #111, opened 2020-04-16
[WIP] Add terra-rstudio-pro-base

rtitle:
Creating this PR as I'm testing RStudio Pro licensing through Leo. It seems to work with this image.

TODO:
1. Update documentation and CI
2. Determine if anything else should be added to the base image (more Bioc setup?)
3. Review with security team -- is using `rstudio/rstudio-server-pro:1.2.5042-1` as a base image OK?


rtitle:
@nturaga you might have seen this PR. It's not ready to merge, but I opened it as a POC to prove that RStudio Pro licensing works through Leo (it does).

I think we should do the same Bioc setup that you've done for the other R images. Your input would be welcome here.

The other major thing to do is work with Albano from a security perspective -- e.g. not sure if we should use a different base image.

This work is tracked in https://broadworkbench.atlassian.net/browse/IA-1815 btw

nturaga:
Hi @rtitle 

I've seen the PR. It's great that we have a working version of this. I will work on a PR of "this" branch to build an RStudio pro bioconductor image.

I'll get back to you shortly.

rtitle:
Thanks, that'd be great!

I think this image is intended to eventually replace the https://github.com/anvilproject/anvil-docker repo btw -- since we have licensing in place we can put all RStudio images in the terra-docker repo.

adrazhi:
> https://broadworkbench.atlassian.net/browse/IA-1815

@rtitle I run an analysis ([see it here](https://gist.githubusercontent.com/adrazhi/82af0d98559bbd5ed0fb6a1e92acd091/raw/d10fbaf37f4b5cbd8e48995b86cdb8cd1ce39fa8/rstudio-server-pro:1.2.5042-1.json)), and the `rstudio/rstudio-server-pro:1.2.5042-1` does have some vulnerabilities, although not particularly scary considering what's out there in the docker world. 

I was wondering if starting off with something like [this](https://github.com/GoogleContainerTools/distroless#docker) would be possible.

rtitle:
> I was wondering if starting off with something like this would be possible.

Cool yeah I bet distroless images is possible here -- can look into it.

PR #113, opened 2020-04-17
IA-1856 Compile pysam explicitly with google cloud support

gokceneraslan:
Google bucket support of pysam is not enabled by default in the PyPI package. It would be great to enable it explicitly in the docker file.

### DIFF
--- a/terra-jupyter-python/Dockerfile
+++ b/terra-jupyter-python/Dockerfile
@@ -80,7 +80,7 @@ RUN pip3 -V \
  && pip3 install wheel
 
 ENV HTSLIB_CONFIGURE_OPTIONS="--enable-plugins --enable-libcurl --enable-gcs"
-pip3 install pysam==0.15.4 --no-binary pysam
+pip3 install Cython && pip3 install pysam==0.15.4 --no-binary pysam
 
 ENV USER jupyter-user
 USER $USER
--- a/terra-jupyter-python/Dockerfile
+++ b/terra-jupyter-python/Dockerfile
@@ -62,7 +62,6 @@ RUN pip3 -V \
  && pip3 install protobuf==3.7.1 \
  && pip3 install pymc3==3.1 \
  && pip3 install pyparsing==2.2.0 \
- && pip3 install pysam==0.15.1 \
  && pip3 install python-dateutil==2.6.1 \
  && pip3 install pytz==2017.3 \
  && pip3 install pyvcf==0.6.8 \
@@ -79,6 +78,9 @@ RUN pip3 -V \
  && pip3 install readline==6.2 \
  && pip3 install setuptools==42.0.2 \
  && pip3 install wheel
+
+ENV HTSLIB_CONFIGURE_OPTIONS="--enable-plugins --enable-libcurl --enable-gcs”
+pip3 install pysam==0.15.4 --no-binary pysam
 
 ENV USER jupyter-user
 USER $USER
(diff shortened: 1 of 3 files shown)


Qi77Qi:
Hi @gokceneraslan , thanks for contributing! Can you update the image version following https://github.com/DataBiosphere/terra-docker#generate-new-image as well (last 2 bullet points probably don't apply)...If you can test the image manually, that'd be great; if not, that's okay too.

gokceneraslan:
Updated both json file and the markdown file. I cannot test it right now but I think Jason Cerrato tested that the notebook works fine. But nobody tested if the gcloud feature works.

gokceneraslan:
OK, the last version is confirmed to be working. Would be great if you could merge it. Thanks.

Qi77Qi:
@gokceneraslan we plan to assist with this PR soonish. Quick question for you, are you using terra-jupyter-python image explicitly or you're using one of the images derived from it?

gokceneraslan:
We are using terra-jupyter-python Dockerfile + changes in this PR, but there are more people who are looking for this feature in our lab (Regevlab)

gokceneraslan:
But I think it'd be better to apply same changes to terra-jupyter-bioconductor image too, because there are also people using both R and Python.

terra-jupyter-python/Dockerfile
```diff
@@ -80,6 +80,9 @@ RUN pip3 -V \
  && pip3 install setuptools==42.0.2 \
  && pip3 install wheel
 
+ENV HTSLIB_CONFIGURE_OPTIONS="--enable-gcs"
+pip3 install Cython && pip3 install pysam==0.15.4 --no-binary pysam
```
rtitle, line 84: I think this needs to be prefixed with `RUN`?

Actually, can we put the `ENV` up above and chain this `pip3 install` in the existing layer? Something like:
```
ENV HTSLIB_CONFIGURE_OPTIONS="--enable-gcs"

RUN pip3 -V \
 && pip3 install --upgrade pip \
 … all the stuff …
 && pip3 install Cython \
 && pip3 install pysam==0.15.4 --no-binary
```

rtitle:
Left a comment but overall I like the change, thanks for PR-ing. Might make sense to merge this PR first and then make a follow-on PR for updating the derived images (hail, aou, gatk, bioconductor). Our team can probably do that.

TamaraO:
I just wanted to confirm that the latest PR from @gokceneraslan works for me in the Terra Notebook and is a very useful feature that would be helpful for anyone using pysam to read BAM files. It does not make sense to copy BAM files locally first when they can be accessed directly. 

One caveat is that to access protected files, GCS token must be generated. Adding this line to my Notebook works:
`os.environ.putenv('GCS_OAUTH_TOKEN', subprocess.check_output('gcloud auth application-default print-access-token', shell=True))`

It seems to remain active for some short period of time and then needs to be done again, and I am not sure if there is a workaround for this yet.

terra-jupyter-python/Dockerfile
gokceneraslan: Sounds good. Yes, sorry about the missing RUN. Apparently people tested a slightly different dockerfile without this mistake 😓

gokceneraslan: OK, I think I addressed that now. Thanks!

gokceneraslan:
> Left a comment but overall I like the change, thanks for PR-ing. Might make sense to merge this PR first and then make a follow-on PR for updating the derived images (hail, aou, gatk, bioconductor). Our team can probably do that.

Sounds good, I think it's ready to me merged.

PR #115, opened 2020-04-22
terra-jupyter-r:0.0.12

nturaga:
- Create new entry in CHANGELOG file.

- This image will include most of CRAN and Bioconductor system dependencies.

- Makes the notebook actually useable by analysts.

- Remove Bioconductor packages

- Add environment variables R_PLATFORM and R_PLATFORM_BINARY_VERSION, these are needed to make the binary package installation of R packages.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -73,7 +73,7 @@
             "base_label": "R",
             "tools": ["r"],
             "packages": {},
-            "version": "0.0.11",
+            "version": "0.0.12",
             "automated_flags": {
                 "include_in_ui": false,
                 "generate_docs": true,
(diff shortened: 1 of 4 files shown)


Qi77Qi:
https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json needs to be updated as well..also are you planning on updating bioconductor image, gatk image and aou image as well to use the new R image?

nturaga:
> https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json needs to be updated as well..also are you planning on updating bioconductor image, gatk image and aou image as well to use the new R image?

I'll update this. 

And yes, i'll update the bioconductor image, gatk image and aou image as well. It would be simply changing the version numbers in the gatk and aou images. The bioconductor image will come with some significant differences.

nturaga:
@Qi77Qi Should I put them all in this PR?

Qi77Qi:
@nturaga maybe better to put them in a separate PR just so we can have this change go through automation tests first and validated?

PR #118, opened 2020-04-24
IA-1867 Downstream fixes for image inherited from terra-jupyter-r

nturaga:
- These images inherit from terra-jupyter-r:0.0.13, the changes made
bump the version number of each of the images.

- terra-jupyter-bioconductor:0.0.13 contains removal of redundant
installations due to the update in terra-jupyter-r:0.0.12 image.

- It's possible that there are redundant dependencies in
terra-jupyter-aou and terra-jupyter-gatk as well. These should be
handled by members who have more knowledge of the images, by comparing
to the dependencies installed in the terra-jupyter-r image.

- I saw no reason to decouple the PR as it's simply a version bump in most of them. I've tested the `bioconductor` image and it should be good to go. If it fails any tests on the CI then, i'll make a separate PR, but I expect nothing to change.


Qi77Qi:
@calbach @jmthibault79  just want to cc you guys about this change. https://github.com/DataBiosphere/terra-docker/pull/115 is the previous PR that brought some changes to base R image...The gist of the change is some system dependencies change for R (I mentioned about it in slack few days ago)...If you guys would like to hold off on bringing in these changes, we can leave aou image out in this PR

calbach:
@Qi77Qi thanks for the fix - sounds good to bump the AoU version, we probably won't switch this before launch, but no harm in having it ready to test out.

PR #119, opened 2020-04-28
Fix bug in bioconductor image

rtitle:
All images need to launch as `jupyter-user`, not as `root`. Will run Leo automation tests against this.


nturaga:
Hi @rtitle, Thanks for getting to this. I was just about to look into why these issues.

PR #123, opened 2020-05-15
Update install_jupyter_contrib_nbextensions.sh

Marianie-Simeon:
Overview:
Added code to enable the Codefolding and Collapsible_header extension to the base image using the install_jupyter_contrib_nbextensions.sh

Background:
Many of CD notebooks (GATK workshop notebooks, as well as many template notebooks) include long code functions that users don’t usually (or ever) need to modify. Codefolding could satisfy two distinct audiences - help users without a lot of coding experience by tucking all the nitty gritty details of defining functions away-  while allowing people who want to see the code to do so.

CD use these extensions a lot in tutorials and template and enabling them as default would keep us from having to walk users through the process of enabling these extensions.

Enabling codefolding and collapsible sections extensions for everyone would not clutter up existing notebooks, or those of people who didn’t want to use the extensions. Cells are un-folded and un-collapsed by default, even if the extension is enabled) and the pale grey arrows that exist when they’re enabled are unobtrusive.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -59,7 +59,7 @@
             "base_label": "Base",
             "tools": ["python"],
             "packages": {},
-            "version": "0.0.9",
+            "version": "0.0.10",
             "automated_flags": {
                 "generate_docs": true,
                 "include_in_ui": false,
--- a/config/conf.json
+++ b/config/conf.json
@@ -17,7 +17,7 @@
             "base_label": "Bioconductor",
             "tools": ["python", "r"],
             "packages": { "r": ["BiocVersion", "tidyverse"] },
-            "version": "0.0.14",
+            "version": "0.0.15",
             "automated_flags": {
                 "generate_docs": true,
                 "include_in_ui": true,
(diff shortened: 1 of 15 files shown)


Qi77Qi:
@Marianie-Simeon the change looks fine to me...but you'll need to update https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/CHANGELOG.md and https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L62 to bump the image the version (see https://github.com/DataBiosphere/terra-docker/pull/121 as an example). Also since you're changing base image, so you'll probably need to bump all derived images.

Marianie-Simeon:
> @Marianie-Simeon the change looks fine to me...but you'll need to update https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/CHANGELOG.md and https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L62 to bump the image the version (see #121 as an example). Also since you're changing base image, so you'll probably need to bump all derived images.

rtitle:
Sorry just seeing this… the change to the base image looks good, but there are some more changes needed to update the derived images. Do you mind if I add a commit to this PR?

rtitle:
^ this latest commit should correctly update all derived images

Marianie-Simeon:
Thank you!!

Marianie-Simeon:
Also, I don't mind! Thank you Rob!

rtitle:
Cool - so I think I will do a quick test, then thumb and merge this.

Marianie-Simeon:
Sounds Great!

PR #125, opened 2020-05-22
 label change for Legacy GATK image

rtitle:
@sharmatime wanted to add this text. Need to fill in the date when the release date is known.

Opening this PR as a placeholder so I don't forget.

### DIFF
--- a/scripts/generate_version_docs.py
+++ b/scripts/generate_version_docs.py
@@ -109,7 +109,7 @@ def get_static_docs():
     },
     {
       "id": 'terra-jupyter-gatk_legacy',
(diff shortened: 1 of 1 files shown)


scripts/generate_version_docs.py
```diff
@@ -109,7 +109,7 @@ def get_static_docs():
     },
     {
       "id": 'terra-jupyter-gatk_legacy',
-      "label": 'Legacy GATK (GATK 4.1.4.1, Python 3.7.7, R 3.6.3)',
+      "label": 'Legacy GATK (default prior to May XX, 2020) (GATK 4.1.4.1, Python 3.7.7, R 3.6.3)',
```
Qi77Qi, line 112: I wonder if we should use a shorter date format since the label is already a bit long

rtitle:
Think this is probably OK to merge now, since our planned release date is June 1

PR #128, opened 2020-05-29
Minor edits to Dockerfile in R package installation

nturaga:
Install "remotes" and "devtools" with BiocManager::install()

### DIFF
--- a/terra-rstudio-pro-base/Dockerfile
+++ b/terra-rstudio-pro-base/Dockerfile
@@ -15,12 +15,12 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
 
 # install BiocManager and google-cloud R packages
 # TODO: should we include other bioconductor setup?
-RUN R -e "install.packages(c( \
-  'remotes', \
-  'BiocManager', \
-  'devtools'), \
-  repos = 'http://cran.us.r-project.org')" \ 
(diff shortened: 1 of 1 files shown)


Qi77Qi:
will this cause any version change?

nturaga:
> will this cause any version change?

No, there will be no version changes. This is just a small modification to follow best practices to have the installation be done in the right places.

nturaga:
Although the R version in the `rstudio/rstudio-server-pro:1.2.5042-1` image which this inherits from is outdated and is R-3.6.3. 

Ideally we would get R-4.0 which is the current release version of R. @rtitle Can you comment on this? Does RStudio have a newer image?

rtitle:
> will this cause any version change?
We haven't started versioning RStudio Pro yet; need to set that up and push an initial `0.0.1` version

rtitle:
> Does RStudio have a newer image?

That's a great question. I don't see R 4.0.0 support in https://github.com/rstudio/rstudio-docker-products. This might be a good question for the RSP support team, I could make a case and cc you

nturaga:
Thanks @rtitle , that would be great to be cc-ed on it.

Qi77Qi:
think this PR is outdated....going to close...feel free to reopen if I missed sth

PR #129, opened 2020-05-29
Add work-in-progress RStudio-pro-bioconductor image

nturaga:
- Add Dockerfile, with sys dependencies installed.

- NOTES: IMO, these sys dependencies should just be in the base
  image just like in terra-jupyter-r, because these sys dependencies
  allow installation and use of most CRAN and Bioconductor packages.

- R version in this image is out of date with version R.3.6.3. The base
  image needs to be built on a more recent version of R.

- Once R-4.0 is used, it's possible to remove installation of
  Bioconductor packages manually as binary package insallation will be
  available with the `AnVIL` package. There will be no need of a special
  Bioconductor image. It can just be `rstudio-pro-base:<version>`, which
  all users use across the board.

cc-ing @rtitle and @mtmorgan for a broader discussion.


mtmorgan:
- yes, at some point (not immediately) it seems like a bioconductor-flavored image could be redundant with a 'most of' CRAN / Bioc image

- It doesn't make sense to make this available as R-3.63; R-4.0.1 is due soon (today?)

- I was wondering about the additional layer RStudioPro presents to the user -- instead of getting an RStudio session, they get a page where they can manage several RStudio sessions. This seems quite confusing, including how this will work with AnVIL cloud resources.

PR #133, opened 2020-06-29
IA-1973 Persist only user installed packages

gpcarr:
This PR removes `/home/jupyter-user/notebooks/packages` from `.Renviron` in `terra-jupyter-r` so that derived images such as `terra-jupyter-bioconductor` package installations don't persist

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -73,7 +73,7 @@
             "base_label": "R",
             "tools": ["r"],
             "packages": {},
-            "version": "1.0.2",
+            "version": "1.0.3",
             "automated_flags": {
                 "include_in_ui": false,
                 "generate_docs": true,
--- a/config/conf.json
+++ b/config/conf.json
@@ -17,7 +17,7 @@
             "base_label": "R / Bioconductor",
             "tools": ["python", "r"],
             "packages": { "r": ["BiocVersion", "tidyverse"] },
-            "version": "1.0.2",
+            "version": "1.0.3",
             "automated_flags": {
                 "generate_docs": true,
                 "include_in_ui": true,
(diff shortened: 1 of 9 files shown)


config/conf.json
```diff
@@ -73,7 +73,7 @@
             "base_label": "R",
             "tools": ["r"],
             "packages": {},
-            "version": "1.0.2",
+            "version": "1.0.3",
```
rtitle, line 76: I think other versions in this file need to be changed as well for: `terra-jupyter-gatk`, `terra-jupyter-aou`, `terra-jupyter-bioconductor`

terra-jupyter-aou/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.0.3 - 06/29/2020
+
+- Update `terra-jupyter-r` base image to `1.0.3`
```
rtitle, line 3: It'd be nice to put some text in here about the actual change

config/conf.json
gpcarr: oh right, good catch

Qi77Qi: if the previous versions are bad, shall we just remove the bad versions, keep versions the same in this PR?

terra-jupyter-r/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.0.2 - 06/29/2020
```
rtitle, line 1: Should be 1.0.3

config/conf.json
gpcarr: going to merge as is so we have a full history of the changes made to these images

PR #134, opened 2020-06-29
IA-1973 Fix AOU image build failure bug

gpcarr:
built `terra-jupyter-aou` using new test r image and build succeeded


terra-jupyter-r/Dockerfile
```diff
@@ -150,6 +150,9 @@ RUN R -e 'install.packages("BiocManager")' \
     "uuid"))' \
     && chown -R $USER:users /home/jupyter-user/.local
 
+## pip runs as jupyter-user
+ENV PIP_USER=true
```
Qi77Qi, line 154: I think we should wait until aou confirm it's okay to do so

PR #138, opened 2020-07-27
Upgraded hail to version 0.2.50

johnc1231:
Upgrading hail to version 0.2.50.

PR #139, opened 2020-07-27
[IA-2076] fix issue with path and update all images

jdcanas:
Tested by building a custom image, pushing it, and installing a pip package with cli functionality. (`janis-pipelines` to be specific).

Worth noting that I couldn't test this locally with an image because of the directory setup our init script does... may be worth touching up a terra-docker script that sets up the pathing as the init script does.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -17,7 +17,7 @@
             "base_label": "R / Bioconductor",
             "tools": ["python", "r"],
             "packages": { "r": ["BiocVersion", "tidyverse"] },
-            "version": "1.0.3",
+            "version": "1.0.4",
             "automated_flags": {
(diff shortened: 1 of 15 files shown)


config/conf.json
```diff
@@ -31,7 +31,7 @@
             "base_label": "Hail",
             "tools": ["python", "spark"],
             "packages": { "python": ["hail"] },
-            "version": "0.0.15",
+            "version": "0.0.16",
```
rtitle, line 34: Note hail was bumped to `0.0.16` in https://github.com/DataBiosphere/terra-docker/pull/138 -- you may want to let that one merge first them bump it to `0.0.17` here

PR #148, opened 2020-08-26
IA-2040 remove explicit notebook install from python image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2040

We install notebook in base. it shouldn't be needed in python image.
I'll delete the `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.15` from gcr, hence not bumping version

### DIFF
--- a/terra-jupyter-python/Dockerfile
+++ b/terra-jupyter-python/Dockerfile
@@ -30,7 +30,6 @@ RUN pip3 -V \
  && pip3 install pandas-gbq==0.12.0 \
  && pip3 install pandas-profiling==2.4.0 \
  && pip3 install seaborn==0.9.0 \
- && pip3 install notebook==5.7.8 \
  && pip3 install terra-notebook-utils==0.2.1 \
  && pip3 install python-lzo==1.12 \
  && pip3 install google-cloud-bigquery==1.23.1 \


rtitle:
Hmm.. we'll need to update derived images too, right? I think bumping versions and letting CI run is the better way to go.

Qi77Qi:
@rtitle `0.0.15` is a bad version tho...if we bump version, we'll need to bump all image versions, vs if we don't bump, we just need to delete `0.0.15` python image and rebuild it, and the rest will be the same...If we bump versions again, I'll need to make sure https://github.com/DataBiosphere/leonardo/pull/1573 are bumped properly, and the subsequent custom image also have correct versions...It's a lot simpler if all I need to verify is all versions are bumped by 1

rtitle:
My main point was that even if you replace `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.15`, the derived images won't pick up the change until we replace them, e.g.: `terra-jupyter-gatk`, `terra-jupyter-hail`, etc. Was just thinking if we need to rebuild many images it's probably easier to let CI do it. Maybe I'm misunderstanding something?

Qi77Qi:
no, you're definitely right...all images need to be rebuilt except base. I'll still let CI to rebuild all these images after I delete them in GCR...

rtitle:
Oh okay that sounds good

PR #149, opened 2020-09-02
[IA-2169] Terminal default directory is now where PD is mounted

gpcarr:
Add line to `/home/jupyter-user/.bashrc` that defaults the user's terminal to the directory where the PD is mounted.

Tested all of the following images (built them with my test base image with the new change): `terra-jupyter-base`, `terra-jupyter-python`, `terra-jupyter-aou` and verified that when I opened Jupyter then clicked New --> Terminal that the default directory is the correct one:
Screen Shot 2020-09-02 at 12 18 46 PM

I also verified that everything pip that the images install is in the same directory as before (installed as root user and for the aou image some are installed as `jupyter-user` so I made sure those are where they're supposed to be as well)

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -26,7 +26,7 @@
                     "tidyverse"
                 ]
             },
-            "version" : "1.0.5",
+            "version" : "1.0.6",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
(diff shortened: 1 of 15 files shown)


Qi77Qi (approved):

rtitle (approved):
LGTM

PR #158, opened 2020-09-18
AoU image: add Hail dependencies

calbach:
AoU ticket: https://precisionmedicineinitiative.atlassian.net/browse/RW-5412

- Add Hail to AoU image
- Bump GATK branch commit

Verified that Hail is functional with Dataproc, and pyspark fails to import on GCE, which is reasonable behavior for now.

CC @ahaessly @petesantos @ericsong @rtitle @als364

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "1.0.9",
+            "version" : "1.0.10",
             "automated_flags" : {
                 "include_in_ui" : false,
                 "generate_docs" : false,
--- a/terra-jupyter-aou/CHANGELOG.md
+++ b/terra-jupyter-aou/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 1.0.10 - 09/18/2020
 
+ - Add Hail / Spark libs.
  - Bump GATK jar to latest commit ah_var_store branch:
    https://github.com/broadinstitute/gatk/tree/ah_var_store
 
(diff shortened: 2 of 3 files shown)


calbach:
Please hold off on merging this for now - going to check whether I can resolve the PDF download issue as well

rtitle (approved):
It looks good to me, but I defer to AoU for the testing. 

Just as an observation: we used to have a single mega-image for FireCloud/Terra but moved to an image stack for better composability -- e.g. fewer conflicts between tools, smaller images if a user only cares about certain functionality, easier upgrades, etc. I can also see the benefit of an all-in-one image though, and it's interesting to see AoU moving in that direction.

rtitle:
re enabling PDF, check the diff in image size. I recall that installing latex took an unreasonable amount of space, but that was also >1y ago.

rtitle:
Also leaving this to be aware of: https://broadworkbench.atlassian.net/browse/IA-2221

I saw it on Qi's handoff doc, I don't know the details/context.

calbach:
Thanks - I did notice yesterday that I could not pull the latest hail image, so that explains it.

I'm hitting a slow trickle of additional issues with the PDF fix I add more and more libraries.. various additional packages are needed in order to properly render. Going to punt on the PDF issue for now. So far, these dependencies have increased the image size by ~.5GB in my testing, but more packages are still needed.

calbach:
Merging 🤞

PR #159, opened 2020-09-29
AoU: Download gatk jar instead of building

calbach:
This should massively reduce the resource requirements for building this image, per

- https://broadworkbench.atlassian.net/browse/IA-2221
- https://broadworkbench.atlassian.net/browse/DDO-752

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "1.0.10",
+            "version" : "1.0.11",
             "automated_flags" : {
                 "include_in_ui" : false,
                 "generate_docs" : false,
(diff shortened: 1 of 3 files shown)


calbach:
CC @ericsong @ahaessly

rtitle (approved):
Looks reasonable to me - I wonder if we should adopt this in `terra-jupyter-gatk` as well

calbach:
`terra-jupyter-gatk` already downloads the zip from the gatk releases, so it's effectively doing the same thing. I just need to grab it from a different location because this is a branched version of the jar - later this will be merged into the main line and we can use the same approach as `terra-jupyter-gatk`, probably just doing a `COPY` from it.

rtitle:
BTW @calbach -- I suspect you're testing these images yourself, but we can also bump the version [here](https://github.com/DataBiosphere/leonardo/blob/develop/automation/src/test/resources/reference.conf#L5) to run Leo automation tests against the latest and greatest AoU image.

PR #164, opened 2020-10-19
[IA-2255] Add RStudio/Bioconductor image to cloud environments dropdown

gpcarr:
Tested by doing the following:
- ran the `generate_version_docs.py` script, replacing `terra-docker-versions-new.json` with `gabriela-test-rstudio.json` and verified that the json file that gets generated is what we expect it to be, the only difference being that the new file includes the RStudio image.
- The corresponding terra-ui changes can be found in this PR: https://github.com/DataBiosphere/terra-ui/pull/2291
- Tested that I could create a runtime using the new rstudio image on my local terra-ui branch (by using the `gabriela-test-rstudio.json` file that i pushed to the google bucket) and everything is working as expected

Screenshot of terra-ui changes:
Screen Shot 2020-10-21 at 3 25 05 PM

### DIFF
--- a/config/community_images.json
+++ b/config/community_images.json
@@ -10,7 +10,7 @@
 },
 {
     "id": "RStudio",
-    "label": "RStudio / Bioconductor (Python 2.7.17, BiocVersion 3.11.1, tidyverse 4.0.0)",
+    "label": "RStudio (R 4.0.0, Bioconductor 3.11.1, Python 2.7.17)",
     "version": "0.0.6",
     "updated": "2020-07-01",
     "packages": "https://storage.googleapis.com/terra-docker-image-documentation/placeholder.json",
--- a/config/community_images.json
+++ b/config/community_images.json
@@ -7,4 +7,14 @@
     "image": "cumulusprod/pegasus-terra:1.0",
     "requiresSpark": false,
     "isCommunity": true
+},
+{
+    "id": "RStudio",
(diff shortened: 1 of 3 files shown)


Qi77Qi:
can you attach a screen shot of what the new view looks like

config/community_images.json
```diff
@@ -7,4 +7,14 @@
     "image": "cumulusprod/pegasus-terra:1.0",
     "requiresSpark": false,
     "isCommunity": true
+},
+{
+    "id": "RStudio",
+    "label": "RStudio (R 4.0.0, Bioconductor 3.11.1, Python 2.7.17)",
+    "version": "0.0.6",
+    "updated": "2020-07-01",
+    "packages": "https://storage.googleapis.com/terra-docker-image-documentation/placeholder.json",
```
rtitle, line 16: So we should ask the Bioconductor team to generate a list of packages, correct? (I think we talked about this in person)

rtitle (approved):

gpcarr:
> can you attach a screen shot of what the new view looks like...does this depend on UI or the other way around?

I'll attach a screenshot of what the new view looks like in the description section of this PR :) 
We need to release the UI first otherwise the UI will still display the RStudio image but in the wrong section of the dropdown.

config/community_images.json
gpcarr: Yeah, I was planning on letting them know once this is released but now that I think about it, I should probably reach out sooner so that their json file is in place once the release happens

PR #166, opened 2020-10-22
Don't use gsutil in the aou Dockerfile, causes CLI issues

calbach:
Invoking gsutil from the Docker build process has side-effects: avoid.

As an end user, this resulted in errors such as:

```
!gsutil ls ${WORKSPACE_BUCKET}

Traceback (most recent call last):
  File "/usr/lib/google-cloud-sdk/lib/googlecloudsdk/core/util/files.py", line 1227, in _FileOpener
    return io.open(path, mode, encoding=encoding)
PermissionError: [Errno 13] Permission denied: '/home/jupyter-user/.config/gcloud/active_config'
```

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "1.0.13",
+            "version" : "1.0.14",
             "automated_flags" : {
                 "include_in_ui" : false,
                 "generate_docs" : false,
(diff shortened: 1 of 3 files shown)


calbach:
OK... now that should do it.

calbach:
This one's borked, replaced with https://github.com/DataBiosphere/terra-docker/pull/167

PR #174, opened 2020-11-04
[IA-1988] Add sudo support to base image

rtitle:
https://broadworkbench.atlassian.net/browse/IA-1988

Depends on gvisor enablement within Leo. Still needs some more testing.


rtitle:
Closing, will go with a different approach

PR #180, opened 2020-11-16
[IA-2322] Remove miniconda to fix terminal

jdcanas:
The versions match in terminal and in a notebook when tested locally.

### DIFF
--- a/updateVersions.sc
+++ b/updateVersions.sc
@@ -1,6 +1,6 @@
-val workbenchUtil2 = "0.10-c3ef6b80-SNAP"
+// Must be run with scala 2.13 or above.
 
-interp.configureCompiler(_.settings.YpartialUnification.value = true)
+val workbenchUtil2 = "0.10-c3ef6b80-SNAP"
 
 val circeVersion = "0.13.0"
 interp.load.ivy(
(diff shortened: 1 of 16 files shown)


updateVersions.sc
```diff
@@ -1,6 +1,6 @@
-val workbenchUtil2 = "0.10-c3ef6b80-SNAP"
+// Must be run with scala 2.13 or above.
 
-interp.configureCompiler(_.settings.YpartialUnification.value = true)
```
jdcanas, line 3: This line causes the script to fail if you run with scala 2.13. The feature is now enabled by default. https://stackoverflow.com/questions/61035986/value-ypartialunification-is-not-a-member-of-scala-tools-nsc-settings

updateVersions.sc
```diff
@@ -1,6 +1,6 @@
-val workbenchUtil2 = "0.10-c3ef6b80-SNAP"
+// Must be run with scala 2.13 or above.
```
Qi77Qi, line 1: I think comment should be `must be run `ammonite for Scala 2.13+`

Qi77Qi (approved):

rtitle:
I think the aou image build failed as a result of this: https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-build-terra-docker/1528/console

May want to confirm with AoU their use of conda

rtitle:
Also may want to revert in the interim and delete the images that were built in GCR

PR #182, opened 2020-11-17
Drop terra-jupyter-aou dependency on miniconda

calbach:


jdcanas (approved):

rtitle (approved):
LGTM

@jdcanas do we think CI will handle this since we previously bumped `terra-docker-aou`? Or will things still be in a bad state due to the partial build failure? (I'm not sure myself)

jdcanas:
I think it will be good because the CI didn't build the AOU image, so I expect it to build it on this merge and be happy moving forward

calbach:
Merging... lets see.

rtitle:
it looks like `terra-jupyter-aou:1.0.18` got built and pushed to GCR:

https://fc-jenkins.dsp-techops.broadinstitute.org/job/leonardo-build-terra-docker/1582/

PR #183, opened 2020-12-02
Bump terra-notebook-utils to version 0.7.0

xbrianh:

PR #184, opened 2020-12-02
Bump versions + changelog for https://github.com/DataBiosphere/terra-docker/pull/183

gpcarr:
tested changes to `updateVersions.sc` in a separate branch, ran `amm ./updateVersions.sc terra-jupyter-r "bump terra-notebook-utils version to 0.7.0"` and verified that we don't bump the version for `terra-jupyter-python` in any Dockerfile in the `FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.20 AS python` section

bumping versions for https://github.com/DataBiosphere/terra-docker/pull/183

### DIFF
--- a/updateVersions.sc
+++ b/updateVersions.sc
@@ -123,7 +123,7 @@ def main(updatedImage: String, updatedImageReleaseNote: String): Unit = {
                 val firstSplit = s.split(":")
                 val splited = firstSplit(1).split("\\.")
                 s"${firstSplit(0)}:${splited(0)}.${splited(1)}.${splited(2).toInt + 1}\n"
-              } else if(s.contains(" AS ")) {
+              } else if(s.contains(" AS ") && imagesToUpdate.contains("terra-jupyter-python")) {
                 val firstSplit = s.split(":")
                 val secondSplit = firstSplit(1).split(" ")
                 val splited = secondSplit(0).split("\\.")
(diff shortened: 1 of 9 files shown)


updateVersions.sc
```diff
@@ -123,7 +123,7 @@ def main(updatedImage: String, updatedImageReleaseNote: String): Unit = {
                 val firstSplit = s.split(":")
                 val splited = firstSplit(1).split("\\.")
                 s"${firstSplit(0)}:${splited(0)}.${splited(1)}.${splited(2).toInt + 1}\n"
-              } else if(s.contains(" AS ")) {
+              } else if(s.contains(" AS ") && imagesToUpdate.contains("terra-jupyter-python")) {
```
Qi77Qi, line 126: I think this won't be valid if the image getting updated is `terra-jupyter-base`

gpcarr: hmm i have a separate branch where i made a change to `terra-jupyter-base` so I can test there

Qi77Qi: it defnitely won't work because when you change terra-jupyter-base, both terra-jupyer-python and terra-jupyter-r will change...so the original logic is fine in that case. you need to somehow make this conditional depends on whether the "base" image is `terra-jupyter-r`

rtitle:
Shall we close this @gpcarr ?

gpcarr:
> Shall we close this @gpcarr ?

yep, will do so now!

gpcarr:
using https://github.com/DataBiosphere/terra-docker/pull/185 in place of this PR

PR #185, opened 2020-12-02
Move terra-notebook-utils to base image and bump all image versions

gpcarr:
going to be closing this PR: https://github.com/DataBiosphere/terra-docker/pull/184

changes to `updateVersions.sc` were tested using both `amm ./updateVersions.sc terra-jupyter-base "[message]"` and `amm ./updateVersions.sc terra-jupyter-r "[message]"`

### DIFF
--- a/updateVersions.sc
+++ b/updateVersions.sc
@@ -122,7 +122,10 @@ def main(updatedImage: String, updatedImageReleaseNote: String): Unit = {
               if(s.contains("FROM") && !s.contains("AS")) {
                 val firstSplit = s.split(":")
                 val splited = firstSplit(1).split("\\.")
-                s"${firstSplit(0)}:${splited(0)}.${splited(1)}.${splited(2).toInt + 1}\n"
+                val imageName = firstSplit(0).split("\\/")(2)
+                if(imagesToUpdate.contains(imageName)) {
+                  s"${firstSplit(0)}:${splited(0)}.${splited(1)}.${splited(2).toInt + 1}\n"
+                } else s"${s}\n"
               } else if(s.contains(" AS ") && imagesToUpdate.contains("terra-jupyter-python")) {
                 val firstSplit = s.split(":")
                 val secondSplit = firstSplit(1).split(" ")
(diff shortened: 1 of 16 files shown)


updateVersions.sc
```diff
@@ -123,7 +123,7 @@ def main(updatedImage: String, updatedImageReleaseNote: String): Unit = {
                 val firstSplit = s.split(":")
                 val splited = firstSplit(1).split("\\.")
                 s"${firstSplit(0)}:${splited(0)}.${splited(1)}.${splited(2).toInt + 1}\n"
-              } else if(s.contains(" AS ")) {
+              } else if(s.contains(" AS ") && imagesToUpdate.contains("terra-jupyter-python")) {
```
rtitle, line 126: Just thinking through.. this works because we do:
```
FROM r-image
FROM py-image AS python
```
right? So we will only replace the second line if the py image is actually being updated.

This would break if we used "AS" in other ways, like
```
FROM some-other-image AS foo
```

Maybe could change it to:
```
else if (s.endsWith(" AS python") && imagesToUpdate.contains("terra-jupyter-python"))
```
to be a bit more robust?

rtitle (approved):
The version bump + change looks correct, comment in the amm script

updateVersions.sc
gpcarr: correct - it would break if we used anything but `python` after `AS`. I'll update now!

Qi77Qi: does this work if we only want to update R?

rtitle: I think that's what this fixes -- now the else if clause won't pass and we won't mistakenly update python.

gpcarr: yeah I tested with `amm ./updateVersions.sc terra-jupyter-r "[message]"` and the versions bumps looks good as seen in this PR https://github.com/DataBiosphere/terra-docker/pull/184/files

Qi77Qi: oh nvm..I thought we have `AS R` for `FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:??? ` as well

Qi77Qi (approved):

rtitle (approved):
:shipit:

PR #186, opened 2020-12-03
Fix base image

gpcarr:
This PR fixes the base image by adding `pip3 install six==1.15.0` and installing `python3.7-dev` instead of `python3.7` and moving `pip3 install terra-notebook-utils==0.7.0` to after we install `firecloud` because TNU depends on firecloud

Removes `six` and `python3.7-dev` installation from python and r images since now we install them in base image

### DIFF
--- a/terra-jupyter-base/Dockerfile
+++ b/terra-jupyter-base/Dockerfile
@@ -115,7 +115,7 @@ ENV PATH ${PATH:+$PATH:}$USER_INSTALL_DIR/bin
 RUN apt update \
    && apt install -yq software-properties-common \
    && add-apt-repository ppa:deadsnakes/ppa \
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):
the change looks good to me..can you create a ticket capturing the terra-notebook-utils version bump and six version bump? mostly for terra release notes later

rtitle (approved):
Looks good -- as discussed can merge this with no version bumps, since the images in `master` are currently broken

rtitle:
+1 making a ticket for the `tnu` bump for release notes

gpcarr:
> the change looks good to me..can you create a ticket capturing the terra-notebook-utils version bump and six version bump? mostly for terra release notes later

will do!

Issue #175, opened 2020-11-05
Remove obsolete Python package enum34


deflaux:
By default, use the standard library enum module instead. 

For more detail, please see https://stackoverflow.com/a/45716067/4138705

rtitle:
This is merged, it should show up in Terra UI after the release next week.

deflaux:
thanks!

deflaux:
This obsolete Python package was still getting pulled into current images due to a dependency on an old version of pymc3.

PR #188, opened 2020-12-09
Update to a newer version of pymc3.

deflaux:
Newer versions of this library no longer depend on the obsolete enum34 library.

Fixes https://github.com/DataBiosphere/terra-docker/issues/175

### DIFF
--- a/terra-jupyter-python/Dockerfile
+++ b/terra-jupyter-python/Dockerfile
@@ -60,7 +60,7 @@ RUN pip3 -V \
  && pip3 install keras==2.1.6 \
  && pip3 install patsy==0.4.1 \
  && pip3 install protobuf==3.7.1 \
- && pip3 install pymc3==3.1 \
+ && pip3 install pymc3==3.10.0 \
  && pip3 install pyparsing==2.2.0 \
  && pip3 install Cython \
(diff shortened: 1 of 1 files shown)


deflaux:
In general, is there a reason why we are pinned to old releases of Python libraries for terra-jupyter-python? Compare to terra-jupyter-r which installs the most recent versions of the R packages.

In particular, tensorflow and pandas are very old in terra-jupyter-python. Might it make sense to move to a strategy where terra-jupyter-python only pins versions when necessary? (e.g. due to a know problem with a particular version of a package)

rtitle:
Thanks again for contributing @deflaux, I can get this merged in.

Re pinning, it's a double-edged sword: when we don't pin libraries we've had unexpected breakages when trying to make unrelated image changes. So pinning can be better for traceability, but on the other hand things can fall very out of date if we don't update them regularly.

Not sure what the best answer is but maybe your suggestion to pin risky libraries but not others makes the most sense.

rtitle (approved):

rtitle:
@deflaux this should be in Terra prod now, if you create a new runtime

deflaux:
thank you @rtitle !

PR #189, opened 2020-12-09
[IA-2420] Update to a newer version of pymc3

rtitle:
Original PR with the actual change: https://github.com/DataBiosphere/terra-docker/pull/188

This PR bumps the versions of python, hail, gatk, and aou images.

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -108,9 +108,6 @@ RUN pip3 install --upgrade \
   mizani==0.6.0 \
   # Parent image pins tensorflow to an old alpha version. Override here for now.
   tensorflow==2.3.0 \
-  # This version of numpy should already be installed but isn't. See IA-2122.
-  # This line can be removed once that ticket is resolved.
-  numpy==1.19.0 \
   statsmodels==0.10.0rc2 \
   "git+git://github.com/all-of-us/workbench-snippets.git#egg=terra_widgets&subdirectory=py"
 
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -108,6 +108,7 @@ RUN pip3 install --upgrade \
   mizani==0.6.0 \
   # Parent image pins tensorflow to an old alpha version. Override here for now.
   tensorflow==2.3.0 \
+  numpy==1.18.5 \
   statsmodels==0.10.0rc2 \
   "git+git://github.com/all-of-us/workbench-snippets.git#egg=terra_widgets&subdirectory=py"
 
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -42,7 +42,7 @@ RUN mkdir -p /etc/spark/conf.dist && mkdir -p /etc/hadoop/conf.empty && mkdir -p
     && update-alternatives --install /etc/hadoop/conf hadoop-conf /etc/hadoop/conf.empty 100 \
     && update-alternatives --install /etc/hive/conf hive-conf /etc/hive/conf.dist 100
 
-ENV HAIL_VERSION=0.2.57
+ENV HAIL_VERSION=0.2.61
 ENV PIP_USER=false
 
 # For dataproc clusters, this path with will be automatically mounted. Else,
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.21 AS python
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-python:0.0.22 AS python
 
 FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:1.0.11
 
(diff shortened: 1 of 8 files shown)


terra-jupyter-aou/Dockerfile
```diff
@@ -108,9 +108,6 @@ RUN pip3 install --upgrade \
   mizani==0.6.0 \
   # Parent image pins tensorflow to an old alpha version. Override here for now.
   tensorflow==2.3.0 \
-  # This version of numpy should already be installed but isn't. See IA-2122.
```
rtitle, line 111: @calbach thoughts on this? Since pip changed their default behavior I now see a consistent numpy version in the aou image:
```
jupyter-user@1ea5a0456f81:~$ pip3 show numpy
Name: numpy
Version: 1.18.5
Summary: NumPy is the fundamental package for array computing with Python.
Home-page: https://www.numpy.org
Author: Travis E. Oliphant et al.
Author-email: None
License: BSD
Location: /home/jupyter-user/.local/lib/python3.7/site-packages
Requires: 
Required-by: tensorflow, tensorboard, statsmodels, scipy, plotnine, opt-einsum, mizani, matplotlib, h5py, xarray, Theano, Theano-PyMC, tb-nightly, seaborn, scikit-learn, pymc3, pyerfa, phik, patsy, pandas, numba, netCDF4, missingno, matplotlib-venn, Keras, Keras-Preprocessing, Keras-Applications, hail, ggplot, fastprogress, fastinterval, cftime, bx-python, bokeh, biopython, astropy, arviz

jupyter-user@1ea5a0456f81:~$ python3
Python 3.7.9 (default, Aug 18 2020, 06:22:45) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.__version__
'1.18.5'
```

It doesn't seem like `numpy==1.19.0` is possible because `tensorflow==2.3.0` depends on `<1.19.0`. Before this change, the docker build fails with:
```
ERROR: Cannot install mizani==0.6.0, numpy==1.19.0, plotnine==0.6.0, statsmodels==0.10.0rc2 and tensorflow==2.3.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy==1.19.0
    mizani 0.6.0 depends on numpy
    plotnine 0.6.0 depends on numpy>=1.16.0
    statsmodels 0.10.0rc2 depends on numpy>=1.11
    tensorflow 2.3.0 depends on numpy<1.19.0 and >=1.16.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

Feedback welcome here.

gpcarr: trying to understand the problem - before this change was there a different version of `tensorflow` that did not require a lower version of `numpy`? 

and by removing this line, we should have numpy version `1.15.2` (that's what we install in the python image https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-python/Dockerfile#L25) however you're getting numpy version `1.18.5` which is probably installed as a dependency for a different package. my only concern here is that we're not explicitly installing numpy version `1.18.5` and so this could break in the future since that version is probably being installed by another package. maybe we can instead of removing line 113, replace it with `numpy==1.18.5 \`? just thinking out loud

calbach: Sorry, I'll dig up what the issue was here after my next meeting. I believe we encountered a downstream as a result of having an improper version of numpy installed. However, I'm not sure specifically that 1.19 was required, or if that was arbitrary.

rtitle: Thanks @calbach, will monitor this ticket.

@gpcarr the build is currently broken in master. I don't think the tensorflow version changed, so not sure how it worked before. I agree if we want `1.18.5` it'd be better to explicitly `pip install numpy==1.18.5` than rely on transitive dependencies -- will make that change.

terra-jupyter-aou/Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN mkdir -p /etc/spark/conf.dist && mkdir -p /etc/hadoop/conf.empty && mkdir -p
     && update-alternatives --install /etc/hadoop/conf hadoop-conf /etc/hadoop/conf.empty 100 \
     && update-alternatives --install /etc/hive/conf hive-conf /etc/hive/conf.dist 100
 
-ENV HAIL_VERSION=0.2.57
+ENV HAIL_VERSION=0.2.61
```
rtitle, line 45: I squeezed in a hail upgrade too to match `terra-jupyter-hail` - let me know if you're not ok with this. :) I built and ran the image locally and did some minimal testing

terra-jupyter-aou/Dockerfile
calbach: Yeah, the specific issue was:


```
import pandas_profiling

ImportError: Numpy version 1.16.0 or later must be installed to use Astropy. 
```

I think we're happy with any version above 1.16.0, and it doesn't necessarily need to be explicit, so long as the phantom Python versions issue has now been resolved and we have a deterministic value. I'm fine with hardcoding 1.18.5 as suggested.

terra-jupyter-aou/Dockerfile
calbach: Sounds good - thank you.

calbach (approved):

PR #193, opened 2021-01-19
[IA-2472] Turn on debug-level logging when JUPYTER_DEBUG_LOGGING env var is true

gjuggler:
(Note: see https://github.com/DataBiosphere/leonardo/pull/1833 for related Leo PR keeping both files in sync.)

Turns on conditional debug-level logging for the Jupyter application, controlled by an environment variable.

### DIFF
--- a/terra-jupyter-base/jupyter_notebook_config.py
+++ b/terra-jupyter-base/jupyter_notebook_config.py
@@ -6,6 +6,10 @@
 import os
 
 c = get_config()
+
+if os.environ.get('JUPYTER_DEBUG_LOGGING') == 'true':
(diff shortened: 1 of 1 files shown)


Qi77Qi (approved):
lgtm...I don't thikn the file in leo is used...we should probably remove that

rtitle (approved):
+1 looks good, we'll need to run the script to bump all the versions (can be in a separate PR).

Agreed the Leo version is deprecated, it's there to support old images (may even be obsolete at this point).

rtitle:
Made "version bump" PR here: https://github.com/DataBiosphere/terra-docker/pull/194

PR #195, opened 2021-01-20
IA-2343 use blessed image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2343

1. use blessed image
2. update script to support bumping major version

### DIFF
--- a/terra-jupyter-base/Dockerfile
+++ b/terra-jupyter-base/Dockerfile
@@ -1,6 +1,6 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-FROM marketplace.gcr.io/google/ubuntu1804:latest
+FROM us.gcr.io/broad-dsp-gcr-public/base/python:debian
 USER root
 
 #######################
--- a/config/conf.json
+++ b/config/conf.json
@@ -26,7 +26,7 @@
                     "tidyverse"
                 ]
             },
-            "version" : "1.0.12",
+            "version" : "2.0.12",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
@@ -47,7 +47,7 @@
                     "hail"
                 ]
             },
-            "version" : "0.0.30",
+            "version" : "1.0.30",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
@@ -68,7 +68,7 @@
                     "scikit-learn"
                 ]
             },
-            "version" : "0.0.23",
+            "version" : "1.0.23",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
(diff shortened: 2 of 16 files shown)


terra-jupyter-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-FROM marketplace.gcr.io/google/ubuntu1804:latest
+FROM us.gcr.io/broad-dsp-gcr-public/base/python:debian
```
rtitle, line 3: Does this change from Ubuntu to Debian?

Qi77Qi: yea...debian is the closest distro among [blessed images](https://github.com/broadinstitute/dsp-appsec-blessed-images#dsp-appsec-blessed-base-images)

rtitle:
I'm a little nervous to bump the base image for _all_ Jupyter images TBH -- before merging we should build all images and do a lot of testing.

Is there a strong reason we don't want to use `marketplace.gcr.io/google/ubuntu*` images @dinvlad? I think we chose a Google-maintained base image based on a recommendation.. can't remember from who exactly, could try to dig it up.

Also is there a "blessed" Ubuntu image we could use?

rtitle:
Also: does `terra-docker` fall under the AppSec requirement of "blessed" images? These images don't run in Terra infrastructure per se, but are purely launched/managed/access controlled by Leo in users' Terra billing projects. Just wondering what the actual requirement is.

dinvlad:
@rtitle good point here! we don't have a _strict_ requirement for scanning end-user images; however, since those images are used universally by users throughout Terra, _and_ they do run user-facing _services_, they seemed like a good candidate to scan. So if we scan them, this will raise fewer questions from the auditors later on 😉

dinvlad:
Also re `marketplace.gcr.io/google/ubuntu*` or any other base images, you're free to use those! The only issue with "Marketplace" GCR images is that access to them is granted to `allAuthenticatedUsers`, instead of `allUsers`. This means that one has to have a Google identity to authenticate with GCR first, in order to pull them. It's not possible to pull them anonymously, like the regular `Dockerfile` inside our GitHub workflow does. So to access those, we'd need to supply a Service Account key to the workflow. If that would work for you, we can do that.

rtitle:
Makes sense re scanning, thanks.

I'm not necessarily married to the GCP Marketplace images, but since these are user-facing images I would just like to avoid a lot of headache involved in swapping out the OS/base image. E.g. who knows if this would break some user's code by changing some python/system dependency. I think another reason we used `marketplace.gcr.io/google/ubuntu1804:latest` is for GCP to manage security patches/etc for us. Not sure what trivy thinks though. :)

dinvlad:
Yep totally understandable! So let's sync offline how you could set up a dedicated SA key for this, and I'll help with the setup in the workflow (already tested the Yaml config for this, to be added)

dinvlad:
After thinking a bit more about this use case, you're absolutely right @rtitle that it's not worth the additional effort to set up scanning of these images for user-facing containers, as the impact would be low in our threat model anyways (unlike for regular services).

Please feel free to reverse the PR to remove Trivy from this repo!

rtitle:
Ok great, thanks @dinvlad !

Qi77Qi:
hmm..I was going to update this branch

PR #202, opened 2021-02-05
Update hail to 0.2.62

johnc1231:
This PR updates `terra-jupyter-hail` to use hail version 0.2.62. 

@Qi77Qi  asked if we could update hail to depend on a newer version of pandas. Hail version 0.2.62 makes this change, allowing pandas versions: pandas>=1.1.0,<1.1.5

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -47,7 +47,7 @@
                     "hail"
                 ]
             },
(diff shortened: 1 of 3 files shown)


Qi77Qi (approved):
thanks @johnc1231 !

PR #203, opened 2021-02-09
[IA-2511] Attempt to fix Seurat installation

rtitle:
https://broadworkbench.atlassian.net/browse/IA-2511

### DIFF
--- a/terra-jupyter-r/Dockerfile
+++ b/terra-jupyter-r/Dockerfile
@@ -84,6 +84,7 @@ RUN apt-get update \
 	xfonts-75dpi \
 	biber \
     && ln -s /usr/lib/gcc/x86_64-linux-gnu/7/libgfortran.so /usr/lib/x86_64-linux-gnu/libgfortran.so \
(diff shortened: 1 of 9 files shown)


Qi77Qi (approved):
thanks!

rtitle:
Still testing… also needs jira and image bumps, will update if it works :)

terra-jupyter-r/Dockerfile
```diff
@@ -131,6 +132,7 @@ RUN R -e 'install.packages("BiocManager")' \
     "nlme", \
     "nnet", \
     "rpart", \
+    "Seurat", \
```
rtitle, line 135: Not actually sure if `Seurat` should be in this image, or if we should leave it out (but allow it to be user-installable). Maybe will see how much it adds to the image

rtitle: It does increase the size but doesn't seem too bad:
```
seurat                                              latest    0afa8947e20f   2 hours ago     5.22GB
us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r      1.0.12    f472483aa606   2 weeks ago     4.64GB
```

Thinking of leaving it here -- it does seem to be a popular too for genomics.

PR #205, opened 2021-03-01
Add more smoke tests.

deflaux:
Add smoke test workflows for the GATK and Hail images which also depend on `terra-jupyter-python`. (A subsequent PR will modify the Dockerfile for terra-jupyter-python and add more test coverage.) 

Please see these successful runs of these new workflows:
* https://github.com/deflaux/terra-docker/actions/runs/612153437
* https://github.com/deflaux/terra-docker/actions/runs/612153436

### DIFF
--- a/.github/workflows/test-terra-jupyter-aou.yml
+++ b/.github/workflows/test-terra-jupyter-aou.yml
@@ -15,8 +15,18 @@ on:
     paths:
     - 'terra-jupyter-aou/**'
     - '.github/workflows/test-terra-jupyter-aou.yml'
-    # Note: secrets are not passed to pull requests from forks, so the dev team will need to use the manual workflow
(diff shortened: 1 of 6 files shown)


deflaux:
Please note that these workflows fail here in upstream because GitHub Actions do not pass secrets to pull requests from forks. But these workflows do pass as you can see from the runs in my fork.

deflaux:
Per offline discussion with @Qi77Qi , I updated the comments for what to do when a pull request comes in from a community member since GitHub secrets are not passed to forks. I also installed a workflow trigger for test branches created in upstream for testing of code from forks.

Qi77Qi (approved):

gpcarr (approved):
lgtm 👍

PR #216, opened 2021-05-10
IA-2731 remove legacy image

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2731


rtitle (approved):
Do we cache this in the VM image / boot disk snapshot? (I'm guessing not)

Qi77Qi:
@rtitle I think we removed it a while ago if I remember correctly...but I'll double check

PR #217, opened 2021-05-24
[RW-6633] Add bcftools to All of Us

ericsong:


terra-jupyter-aou/Dockerfile
```diff
@@ -110,6 +110,15 @@ RUN curl -SL --output dotnet.tar.gz "https://dotnetcli.azureedge.net/dotnet/Runt
 
 RUN echo "Sys.setenv(RETICULATE_PYTHON = '$(which python3)')" >> ~/.Rprofile
 
+RUN wget https://github.com/samtools/bcftools/releases/download/1.12/bcftools-1.12.tar.bz2 && \
```
rtitle, line 113: No strong preference really, but wondering if this should be done from a tmp directory like `/tmp/bcftools`?  (Or maybe `/usr/local/share`?)

calbach: +1 (see other tmpdir handling above for examples)

terra-jupyter-aou/Dockerfile
```diff
@@ -110,6 +110,15 @@ RUN curl -SL --output dotnet.tar.gz "https://dotnetcli.azureedge.net/dotnet/Runt
 
 RUN echo "Sys.setenv(RETICULATE_PYTHON = '$(which python3)')" >> ~/.Rprofile
```
calbach, line 111: nit: would probably put this below the bcftools install, just to keep package setup somewhat separate from environment configuration

terra-jupyter-aou/Dockerfile
ericsong: sounds good

calbach (approved):

rtitle:
Change LGTM but looks like the `Test terra-jupyter-aou / test_docker_image` job failed. Do we know if this is a known issue? It would be good to fix the image or the test if needed.

ericsong:
I think it may be unrelated to my changes actually - judging from the same failure on this no op PR https://github.com/DataBiosphere/terra-docker/pull/218

rtitle:
OK - we can look into it separate then I think

rtitle (approved):

ericsong:
I spent the past several hours looking into this but wasn't able to reproduce or make much headway.. I think I'm a little too unfamiliar with the build process to debug it well. At most, I was able to gather that it probably is due to the tensorflow upgrade to 2.5.0. I believe tensorflow packages are compiled with a specific CUDA version in mind and it may be throwing an error as a result. I was surprised to see that tensorflow was installed at all given that we specify tensorflow-cpu as the install target; maybe due to keras pulling in tensorflow as a dependency?

I think I'm going to merge my PR and request that this issue be looked into separately. I believe our image should be OK since it's based off the terra-jupyter-gatk:1.1.2 image which is already built and presumably doesn't have the version incompatibility issue.

PR #219, opened 2021-05-25
Reduce unnecessary package upgrades.

deflaux:
This change is somewhat redundant, but this is the second time terra-widgets dependencies have caused issues with the build for terra-jupyter-aou.

The root cause of the test failures in https://github.com/DataBiosphere/terra-docker/pull/217 will be fixed by https://github.com/all-of-us/workbench-snippets/pull/65

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -133,8 +133,8 @@ RUN R -e 'BiocManager::install(c( \
   "SAIGEgds", \
   "GENESIS"))'
 
-RUN pip3 install --upgrade nbstripout \
+RUN pip3 install \
+  nbstripout \
+  "git+git://github.com/all-of-us/workbench-snippets.git#egg=terra_widgets&subdirectory=py" \
   && mkdir -p /home/$USER/.config/git \
-  && nbstripout --install --global \
-  # The base image already contains the needed dependencies for this package.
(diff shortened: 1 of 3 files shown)


terra-jupyter-aou/Dockerfile
```diff
@@ -133,8 +133,8 @@ RUN R -e 'BiocManager::install(c( \
...
   "GENESIS"))'
 
-RUN pip3 install --upgrade \
-  nbstripout \
-  "git+git://github.com/all-of-us/workbench-snippets.git#egg=terra_widgets&subdirectory=py" \
+RUN pip3 install --upgrade nbstripout \
   && mkdir -p /home/$USER/.config/git \
-  && nbstripout --install --global
+  && nbstripout --install --global \
+  # The base image already contains the needed dependencies for this package.
```
calbach, line 139: This feels like it may be introducing a brittle dependency, you can imagine in the future workbench-snippets might import a package which is not installed in the base image - it seems correct to allow workbench-snippets to define its own package dependencies

Can we instead just remove `--upgrade` from the install line? Would that also resolve the issue?

calbach (commented):
Thanks for sending a fix for this

terra-jupyter-aou/Dockerfile
deflaux: updated!

If later we have some packages that we know we want upgraded, we can add another pip command.

calbach (approved):
Thanks!

rtitle (approved):
Thanks for fixing!

PR #230, opened 2021-06-24
IA-2283: Add gnomad python package to Hail environment

rtitle:
See https://broadworkbench.atlassian.net/browse/IA-2283

This seems to work locally; will also do a test in a real Spark environment in Terra

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -47,7 +47,7 @@
                     "hail"
                 ]
             },
-            "version" : "0.1.2",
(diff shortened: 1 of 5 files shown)


rtitle:
Seems to work in Terra:
image

(Note this command fails locally because it requires `pyspark`.)

Qi77Qi (approved):
👍 thanks for keeping `updateVersions.sc` updated!

PR #243, opened 2021-08-09
GATK with OpenVINO integration with TensorFlow

ravi9:
This will create a new folder `terra-jupyter-gatk-ovtf`. This includes a Dockerfile that allows GATK to run with OpenVINO™ integration with TensorFlow.


lucidtronix (approved):
The CNNScoreVariants side all looks good to me.  Thanks Ravi & Mustafa!

terra-jupyter-gatk-ovtf/gatk-ovtf.patch
```diff
@@ -0,0 +1,118 @@
+From a22f79681d517ba2c6a7677638aab96c09a5aace Mon Sep 17 00:00:00 2001
```
rtitle, line 1: What is this `patch` used for, and is there a reason it needs to be checked in?

rtitle:
Hi, thanks for contributing this! A few questions:

- Is this image currently pushed anywhere? 
- Is there a desire to push it to the same GCR (`us.gcr.io/broad-dsp-gcr-public/...`) as the other images in this repo? If we want to do that then there are some scripts/CI to update, but then it will get tested, built & pushed automatically whenever the base images update. Otherwise it will need to be maintained manually.
- Is there a desire to add this image to the drop-down in Terra UI?
- Alternatively, should we think about folding this into the `terra-jupyter-gatk` image? I am not sure what the exact differences are between this and `terra-docker-gatk`.

I think this should be okay to merge since it's just a new folder, but just want to understand the path forward for this image.

terra-jupyter-gatk-ovtf/gatk-ovtf.patch
cavusmustafa: Hi, below are the changes included in the patch:
- Enabling OpenVINO integration with TensorFlow for CNNScoreVariants inference
- Freezing the Keras model for inference to trim training related layers. (Based on: https://medium.com/@sebastingarcaacosta/how-to-export-a-tensorflow-2-x-keras-model-to-a-frozen-and-optimized-graph-39740846d9eb)
- Additional changes for TF 2.x compatibility

ravi9: Thank you for reviewing @rtitle.
The patch is used to update two python files `vqsr_cnn/inference.py` and `vqsr_cnn/vqsr_cnn/models.py`.

- These minor changes will help in additional performance improvement. 
- It needs to be checked as it will be used when building the docker image. Inside the docker image, this patch will be copied into it and will be used to update the GATK code.

rtitle:
Thanks for the response!

For adding to CI, I think we'd need to update `conf.json`, `Jenkinsfile`, add a `CHANGELOG.md`, and also add some GitHub action unit tests. For adding a new image to the UI it's a small UI update, but I'd want to run that by our PO.

However! We could eliminate that legwork by just adding the `OpenVINO` installation to `terra-jupyter-gatk` and tagging a new version. Is there any clear reason that's a bad idea? If not I'd probably advocate that approach. I can help dig a little more though..

rtitle:
We connected offline and decided to add this as a separate image with separate CI/CD. I'm going to approve and merge this PR as is, and make another PR for update CI scripts and adding test cases.

rtitle (approved):

ravi9:
Hi @xuscrbw, could you let me know few more specifics on your error, so we can better understand your issue. Our Dockerfile is available [HERE](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-gatk-ovtf/Dockerfile) for your reference. Please see [Line-34 ](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-gatk-ovtf/Dockerfile#L34), which shows GATK is installed.

PR #246, opened 2021-08-26
Update hail to 0.2.74

johnc1231:
I made the normal changes that I make when updating hail version in Terra, but it's worth noting that Hail has also updated from using Spark 2.4.x to Spark 3.1.x by default. Are there other changes I have to make to reflect this?

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -47,7 +47,7 @@
                     "hail"
                 ]
             },
-            "version" : "1.0.2",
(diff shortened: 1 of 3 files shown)


johnc1231:
Any updates on this?

johnc1231:
Thanks for the info, that's fine with me.

Qi77Qi (approved):

Issue #244, opened 2021-08-19
GPUs, Terra image builds, and TF >=2.5


amygdala:
Creating this issue for discussion. (I’m on the Terra solutions team at Verily).
It’s a known issue that the Terra images (using the `gcr.io/deeplearning-platform-release/base-cu110:latest` base image) don’t support >=TF 2.5 with GPUs due to a CUDA library mismatch.
I noticed that GCP notebooks, *which use the same base image*, do support >=TF 2.5 with GPUs where the Terra builds do not.

I did a bit of digging and I believe the reason for this is that the `gcr.io/deeplearning-platform-release/tf2-gpu.2-6` images etc. (which use the same base image above) do their own builds of TF from source, via `bazel`, using the CUDA 11.0 libs, rather than doing a `pip install`.    

The TF release notes indicate that from 2.5 onwards, the TF team’s builds (as you’d get from `pip`) are built using the CUDA 11.2 libs.  So, the GCP notebook team’s build from source, using the 11.0 libs, is key to why their images work for TF >=2.5. 
The GCP notebook team does have a bug in to update to using the 11.2 libs, but it’s just a placeholder and may not see movement soon.

We could just document this well and wait for the GCP notebook team to support CUDA 11.2. 
However, I suspect that many people won’t read the docs, will want to use TF >=2.5 and will probably attempt to update via `pip install`  without realizing that it will remove their use of GPUs.  

It might be worth considering **using the [GCP Notebooks images](https://cloud.google.com/deep-learning-containers/docs/choosing-container) as Terra base images**.  These images have many useful ML and data science-related libs installed in addition to the ML frameworks.
If we could get this working it would let us piggyback on the GCP Notebook team’s maintenance and testing of all the libraries that they install on these images, and their dependencies— basically offloading all that work.
To do this this would presumably require overriding their Jupyter config with the Terra Jupyter setup as the last steps in a Terra Dockerfile, probably `terra-jupyter-python`.

Would this approach— using the GCP notebook images rather than their base image as a starting point— be worth trying?
(A related question, that  I haven’t checked yet, is whether there are similar issues with other ML frameworks like PyTorch, but I suspect there are.  We could provide a set of images, based on the GCP notebook images, to choose from.)

/cc @deflaux

amygdala:
One more comment: 
A third option — in addition to documenting the restriction or leveraging the GCP notebook images — would be to build TF from source yourself using the 11.0 libraries, same as the Notebooks team is doing.

[...]

PR #251, opened 2021-09-10
[IA-2902] [IA-2956] Use tf2-gpu.2-6 as base image

rtitle:
See: 
* https://broadworkbench.atlassian.net/browse/IA-2902
* https://github.com/DataBiosphere/terra-docker/issues/244

This PR updates the `terra-jupyter-base` parent image to `gcr.io/deeplearning-platform-release/tf2-gpu.2-6 `. This supports tensorflow 2.6.0 with GPUs, whereas previously only 2.4.0 was supported due to a CUDA library mismatch.

The base image size difference is:
* old: 13.5GB
* new: 16GB
Since we cache all the images I think the size differential is reasonable.

Added a couple test cases around tensorflow which should run on this PR. I'll also do some local manual testing.

Note: this PR also includes https://broadworkbench.atlassian.net/browse/IA-2956

After this PR is merged and the images are built and pushed, we will rebuild the Leo custom image (to update the caches), and then test through Leo before pushing live.

### DIFF
--- a/terra-jupyter-base/Dockerfile
+++ b/terra-jupyter-base/Dockerfile
@@ -1,6 +1,6 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-FROM gcr.io/deeplearning-platform-release/base-cu110:latest
+FROM gcr.io/deeplearning-platform-release/tf2-gpu.2-6 
 USER root
 
 #######################
--- a/terra-jupyter-base/Dockerfile
+++ b/terra-jupyter-base/Dockerfile
@@ -94,8 +94,6 @@ RUN pip3 -V \
  # Hence, make sure to manually test out "launch terminal" button (the button in the green bar next to start and stop buttons)
  # to make sure we don't accidentally break it every time we upgrade notebook version until we figure out an automation test for this
  && pip3 install python-datauri \
- # This is to avoid incompatibility issue, see https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1529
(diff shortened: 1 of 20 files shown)


rtitle:
Looking at tests, I think some images down the stack failed to build… investigating.

Qi77Qi (approved):
lgtm 🎉

Qi77Qi:
shall we update the openvino's verion in dropdown list as well

terra-jupyter-base/Dockerfile
```diff
@@ -94,8 +94,6 @@ RUN pip3 -V \
  # Hence, make sure to manually test out "launch terminal" button (the button in the green bar next to start and stop buttons)
  # to make sure we don't accidentally break it every time we upgrade notebook version until we figure out an automation test for this
  && pip3 install python-datauri \
- # This is to avoid incompatibility issue, see https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1529
- && conda install "nbconvert<6" \
```
rtitle, line 98: `nbcovert<6` was giving me issues with the new base image… the default nbcovert version doesn't seem to have any issues so hopefully this is ok.

terra-jupyter-base/Dockerfile
```diff
@@ -140,6 +138,7 @@ RUN chown -R $USER:users $JUPYTER_HOME \
 
 USER $USER
 EXPOSE $JUPYTER_PORT
+WORKDIR $HOME
```
rtitle, line 141: The new base image's WORKDIR is `/` for some reason.. setting it to `$HOME` which is what it was before.

ravi9:
@rtitle Thank you for updating. Looks like all tests passed. So looks fine.

PR #252, opened 2021-09-13
Fix multipart Jupyter uploads

calbach:
Jupyter's new(?) [default](https://jupyter-notebook.readthedocs.io/en/stable/config.html) file manager is `LargeFileManager`. This supports multipart chunked uploads, which get triggered when a user attempts to upload a file greater than a certain size via the Jupyter UI. Before this change, each chunk would overwrite eachother, resulting in a file which is the tail chunk of the upload.

As a minor optimization, just skip straight to the base manager if we see intermediate chunks.

Also, while I'm here, delete the long-deprecated `DelocalizingContentsManager` (commits are split out, let me know if you prefer separate PRs).


Qi77Qi (approved):
thanks for fixing!

rtitle (approved):
Looks good to me 👍 

This will require a rebuild of all images; however we're doing that anyway with https://github.com/DataBiosphere/terra-docker/pull/251. Shall we merge this first to sneak it in?

calbach:
@rtitle I'll let you click the merge button so I don't interfere with your ongoing operations. From AoU's perspective, would definitely be great to get this fix as soon as we're able.

rtitle:
No worries -- this PR is included in the version bump from the other PR. The CI is running now..

PR #256, opened 2021-09-21
Intel OVTF patch for updating to TF2.6.0/2.5.0

ravi9:
Hi @rtitle,
With the recent update to TF2.5.0, we ran some tests and found that Keras version also needs to match the TF version.
Also, with a newer version of TF, the base GATK is failing, so we had to update the patch as well.

Please note, there is an error in the base GATK as well. We tested GATK-Base-Docker: `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.2`. See [GATK-Base-Error.pdf](https://github.com/DataBiosphere/terra-docker/files/7206539/GATK-Base-Error.pdf). Probably, this is what @xuscrbw was referring to in #243 

The fix is to apply the gatk-ovtf.patch listed in this repo. The fix is in the imports in GATK file:  `../site-packages/vqsr_cnn/vqsr_cnn/models.py`

Also, is there a way to create an additional tag - latest such that we can use `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk-ovtf:latest`? This will be easy to list in the README. :)

Thanks,
Ravi.
cc: @cavusmustafa


rtitle:
Oh, thanks for this @ravi9! Good find. I can try this and also look into apply the patch to `vqsr_cnn/models.py` in the GATK image.

Re `latest` tag, seems that should be easy enough. Will try to update the build script…

ravi9:
@rtitle I just realized the patch has more changes which leads to some more optimizations.
For just fixing the errors in the base image, inside the docker image, I modified this file:
`/home/jupyter/.local/lib/python3.7/site-packages/vqsr_cnn/vqsr_cnn/models.py`

I changed three lines.
- from keras.optimizers import `adam_v2 as Adam`
- cfg = K.tf.`compat.v1.`ConfigProto(intra_op_parallelism_threads=intra_ops, inter_op_parallelism_threads=inter_ops)
- K.set_session(K.`tf.compat.v1.`Session(config=cfg))

Attached is the modified file as well. 

Also, in the base image, the Python package is not installed, not sure if that is by choice.
We need to run the following inside the docker image to get the python package installed.
- `pip install /etc/gatk-4.2.0.0/gatkPythonPackageArchive.zip` 

Test GATK by running the following:
```
gatk CNNScoreVariants \
        -I gs://gatk-tutorials/workshop_2002/2-germline/CNNScoreVariants/bams/g94982_chr20_1m_10m_bamout.bam \
        -V gs://gatk-tutorials/workshop_2002/2-germline/CNNScoreVariants/vcfs/g94982_b37_chr20_1m_15871.vcf.gz \
        -R gs://gcp-public-data--broad-references/hg19/v0/Homo_sapiens_assembly19.fasta \
        -O my_2d_cnn_scored_ovtf.vcf \
        --tensor-type read_tensor \
        --transfer-batch-size 256 \
        --inference-batch-size 256
```

rtitle:
Beautiful, thank you! I was looking at the patch and was thinking along those lines as well.

For now I opened https://github.com/DataBiosphere/terra-docker/pull/257 for the changes to the `ovtf` image (unfortunately the tests only pass on branches not forks). We can make a separate PR for applying the patch to `terra-jupyter-gatk`.

rtitle:
Going to close this in favor of https://github.com/DataBiosphere/terra-docker/pull/257.

ravi9:
Ok great ! Thanks @rtitle

xuscrbw:
@ravi9 For the file fixes in the base image, would we want to apply a patch in the same way you did? Or is there a way to do it without a patch?

rtitle:
@xuscrbw I'll let Ravi confirm, but I suspect we'd want to install gatk in a similar way to the `ovtf` image:
1. Clone the repo
2. Appy the patch
3. Build from source using `gradlew`

This is different from today where we are just downloading a pre-built gatk from GitHub.

 I'm willing to give this approach a shot if it fixes the gatk issues. We should also add some testing for GATK tools to `gatk_smoke_test.ipynb` and `gatk_smoke_test.py`.

ravi9:
@xuscrbw The issue is in the base-gatk-image: `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.2`

I did not create a patch. You can create one based on the updated models.py attached here:
[models.py.txt](https://github.com/DataBiosphere/terra-docker/files/7212150/models.py.txt)

Without the patch, manually you can replace this file directly in 
- `/home/jupyter/.local/lib/python3.7/site-packages/vqsr_cnn/vqsr_cnn/models.py`
- or replace it in `src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/models.py` and then you will need to build GATK. 

Also, you need to run `pip install /etc/gatk-4.2.0.0/gatkPythonPackageArchive.zip`  for the GATK to work with python packages.

ravi9:
@rtitle I believe we can apply the patch directly to the prebuilt gatk zip.
I will try that method and if it works, I will create a PR for the base-gatk.

PR #257, opened 2021-09-22
Intel OVTF patch for updating to TF2.6.0/2.5.0

rtitle:
Pulled in https://github.com/DataBiosphere/terra-docker/pull/256 and added version bumps and `latest` image tag.

Opened from a branch rather than a fork so GitHub actions can run.

cc @ravi9


terra-jupyter-gatk-ovtf/gatk-ovtf.patch
```diff
@@ -1,8 +1,9 @@
-From a22f79681d517ba2c6a7677638aab96c09a5aace Mon Sep 17 00:00:00 2001
```
jdcanas, line 1: is this file intentionally source controlled? Seems to be the output of a `git diff`

jdcanas (approved):
LGTM, I assume the .patch file is something needed I don't have the context for since its alluded to in the PR desc, and the docker changes look good.

terra-jupyter-gatk-ovtf/gatk-ovtf.patch
Qi77Qi: curious as to what this file is for as well :)

Qi77Qi (approved):

terra-jupyter-gatk-ovtf/gatk-ovtf.patch
rtitle: I asked the same thing. :) I think it's a git patch for making gatk work with the latest `tensorflow`. In fact we might do similar in `terra-jupyter-gatk` to fix the issue with `CNNScoreVariants` (see convo in https://github.com/DataBiosphere/terra-docker/pull/256).

cavusmustafa:
Hi @jdcanas, I tried to provide more context about what is inside the patch. If there is anything not clear or anything I can help to explain, please let me know.

The patch includes the changes below on [GATK 4.2.0.0](https://github.com/broadinstitute/gatk/tree/4.2.0.0).

In the file "src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/models.py":
- TF 2.x compatibility
    - Using "tf.compat.v1" for tf 1.x based APIs.
- Fix for TF 2.5/2.6 related issue:
    - Replacing "from keras.optimizers import Adam" with "from keras.optimizers import adam_v2 as Adam".

In the file "src/main/python/org/broadinstitute/hellbender/vqsr_cnn/vqsr_cnn/inference.py":
- Enabling OpenVINO™ integration with Tensorflow
    - Adding two lines of code to enable OVTF
- Additional optimizations
    - Freezing the Keras model and using TF session to execute the inference instead of Keras API (based on ["How to export a TensorFlow 2.x Keras model to a frozen and optimized graph"](https://medium.com/@sebastingarcaacosta/how-to-export-a-tensorflow-2-x-keras-model-to-a-frozen-and-optimized-graph-39740846d9eb)). It helps eliminating training related ops from the model which are not supported by OVTF and results in better performance.
    - If the input data size is not multiple of the specified batch size, the last batch of data might have a smaller batch size than the previous batches (For example, if batch size is 8 but the input data has 18 samples, the last inference iteration will have a batch size of 2). Since the input shape is changed for the last iteration, OVTF will need the re-compile the whole graph with the changed input shape. To prevent this, we add some additional code to fill last batch with fake data (the output calculated with the fake input will be ignored).

rtitle:
@ravi9 @cavusmustafa just FYI the new `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk-ovtf:0.1.2` is in GCR now:

image

https://console.cloud.google.com/gcr/images/broad-dsp-gcr-public/us/terra-jupyter-gatk-ovtf?project=broad-dsp-gcr-public

PR #266, opened 2021-10-19
Add nbdime to AoU image.

deflaux:
nbdime provides better diffs when working with notebooks in source control.

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -45,6 +45,7 @@ ENV PYSPARK_PYTHON=python3
 
 RUN pip3 install pypandoc \
       nbstripout \
+      nbdime \
       papermill \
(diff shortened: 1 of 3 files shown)


terra-jupyter-aou/Dockerfile
```diff
@@ -45,6 +45,7 @@ ENV PYSPARK_PYTHON=python3
 
 RUN pip3 install pypandoc \
       nbstripout \
+      nbdime \
```
calbach, line 48: I'm fine with this - @Qi77Qi @rtitle any interest in pushing this upstream to the python base?

deflaux: If there is interest, I'd be happy to update this PR to make the change there, just let me know!

Both `nbstripout` and `nbdime` should move together.

PR #269, opened 2021-11-08
enable oneDNN optimizatoin by default for TensorFlow

louie-tsai:
Since TensorFlow v2.5, oneDNN optimization is in the release by default, so users could easily leverage the oneDNN optimization by changing an environment variable.
You could refer to below release notes from official TensorFlow v2.5 release.
https://github.com/tensorflow/tensorflow/releases/tag/v2.5.0

This PR is to enable oneDNN optimization via the environment variable and then test it via smoke_test notebook.
Therefore, users could get the best TensorFlow performance on CPU.

For the validation, we add a oneDNN validation section in smoke_test notebook.
image

### DIFF
--- a/terra-jupyter-python/Dockerfile
+++ b/terra-jupyter-python/Dockerfile
@@ -91,6 +91,8 @@ ENV USER jupyter
 USER $USER
 # We want pip to install into the user's dir when the notebook is running.
 ENV PIP_USER=true
+# Enable Intel oneDNN optimizatoin by default
+ENV TF_ENABLE_ONEDNN_OPTS=1
 
 # Note: this entrypoint is provided for running Jupyter independently of Leonardo.
 # When Leonardo deploys this image onto a cluster, the entrypoint is overwritten to enable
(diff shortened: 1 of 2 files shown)


terra-jupyter-python/Dockerfile
```diff
@@ -91,6 +91,8 @@ ENV USER jupyter
 USER $USER
 # We want pip to install into the user's dir when the notebook is running.
 ENV PIP_USER=true
+# Enable Intel oneDNN optimizatoin by default
+ENV TF_ENABLE_ONEDNN_OPTS=1
```
Qi77Qi, line 95: does it make sense to have this in the intel image instead?

louie-tsai: Hi 
Since TensorFlow v2.5, oneDNN optimization is in the release by default, so users could easily leverage the oneDNN optimization by changing an environment variable.
You could refer to below release notes from official TensorFlow v2.5 release.
https://github.com/tensorflow/tensorflow/releases/tag/v2.5.0
image

It doesn't require additional software packages or source codes changes, so it doesn't need to be in the specific image like Intel image.

thanks
Louie

louie-tsai:
moved changes to https://github.com/DataBiosphere/terra-docker/pull/281

Qi77Qi:
I think we can close this now since this is superceded by the tf 1.7 PR

PR #272, opened 2021-11-16
[No ticket] fix script

jdcanas:
Note on lines 42 and 45 we resolve 'doc' into a guaranteed doc, vs remote doc which can be null if no remote docs exist. This should fix any exceptions.


Qi77Qi (approved):
thanks @jdcanas ! tried it locally, and it's working now!

Issue #285, opened 2021-12-20
issues with `pip install` when building library from source for `terra-jupyter-gatk` img 


amygdala:
We're seeing issues with `pip install` for the `terra-jupyter-gatk` image when it's necessary to build the library from source.  

It appears that `PIP_TARGET` is set when `docker exec` is run on the GCE instance: `https://github.com/DataBiosphere/leonardo/search?q=PIP_TARGET`
Whenever we run `pip`, it interprets the environment variable `PIP_TARGET` as the value for the` --home` flag. https://github.com/pypa/pip/issues/8438#issuecomment-674420308
BUT . . . whenever pip tries to install any packages from source, it passes flag `--prefix `and those flags cannot be used together.

By unsetting `PIP_TARGET` for an installation, we can work around the issue. e.g.: 

`!unset PIP_TARGET ; pip install -U docstring-parser==0.13`

Nicole and I created a notebook to demo the issue: https://app.terra.bio/#workspaces/fc-product-demo/try-saturn/notebooks/launch/pip_install_issues.ipynb
The notebook html is attached in case it's useful.
[pip_install_issues.html.zip](https://github.com/DataBiosphere/terra-docker/files/7747532/pip_install_issues.html.zip)

**Other images may have this same issue**, but we haven't tested yet.
Given the workaround, it's not urgent, but other customers are presumably hitting this as well. (I'll send an email with a bit more after the holidays).

/cc @deflaux

deflaux:
I updated our tests to more closely reflect the pip environment we have in production. I then added a test to demonstrate this issue https://github.com/DataBiosphere/terra-docker/pull/286. You can see from the GitHub action test artifacts that the test fails, as expected. I've commented out the final assertion for now, and we can uncomment that assertion after this issue is fixed.

Qi77Qi:
@amygdala @deflaux think this is resolved with https://github.com/DataBiosphere/leonardo/pull/2518....

PR #281, opened 2021-12-14
[IA-3146][IA-3065] Updating to tf 2.7 gpu image

xuscrbw:

PR #286, opened 2021-12-21
Add pip install tests to terra-jupyter-python.

deflaux:
Added new tests to demonstrate:
1) that new Python packages can be installed via pip <-- works
2) that new Python packages **_compiled from source_** can be installed via pip <--- won't work till after https://github.com/DataBiosphere/terra-docker/issues/285 is fixed, the final assertion is commented out for now

Also fixes some test breakage:
* address tensorflow 2.6 test failures by
    * pinning the package version of Keras
    * ensuring that the installation of pytest does not upgrade pyparsing
* fix test auth failures by updating to the latest gcloud github actions

### DIFF
--- a/terra-jupyter-python/tests/smoke_test.ipynb
+++ b/terra-jupyter-python/tests/smoke_test.ipynb
@@ -40,6 +40,108 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Test package installations"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Install a package we do not anticipate already being installed on the base image"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip3 show rich  # Should show not installed."
+   ]
+  },
+  {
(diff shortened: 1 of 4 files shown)


deflaux:
Note that the test failure in terra-jupyter-base has nothing to do with this change - I only added a comment to the Dockerfile. The test failures there will be addressed with the upgrade to tensorflow 2.7 in https://github.com/DataBiosphere/terra-docker/pull/281

Qi77Qi (approved):
this is more for fixing tests...no need to publish new images

terra-jupyter-python/tests/smoke_test.ipynb
```diff
@@ -36,6 +36,120 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "output = !pip3 show docstring-parser\n",
+    "print(output)  # Should show that it is now installed!\n",
+    "# TODO uncomment this test after https://github.com/DataBiosphere/terra-docker/issues/285\n",
```
deflaux, line 148: Uncomment this after https://github.com/DataBiosphere/terra-docker/issues/285 to re-enable the failing test.

PR #287, opened 2021-12-21
Update terra docker image tests to the latest gcloud GitHub action.

deflaux:


deflaux:
Note that the test failure in terra-jupyter-base has nothing to do with this change - its due to a preexisting issue. The terra-jupyter-base test failure will be addressed with the upgrade to tensorflow 2.7 in https://github.com/DataBiosphere/terra-docker/pull/281

Qi77Qi (approved):
thanks @deflaux !

PR #296, opened 2022-02-01
Update hail to version 0.2.83

johnc1231:
Just updating to latest version of hail.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -47,7 +47,7 @@
                     "hail"
                 ]
             },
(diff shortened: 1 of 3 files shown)


Qi77Qi:
thanks @johnc1231 for the PR, can you create the PR from a non fork...just so the tests can run...right now, test won't run becuz the PR is from a fork

johnc1231:
Yes, will close and replace with a non-fork one.

PR #297, opened 2022-02-04
Fix smoke_test.ipynb

raoberman:


Qi77Qi (approved):
@ravi9 thanks for fixing these!

PR #301, opened 2022-02-15
update to hail 0.2.85

danking:


danking:
Hey there friends! We're experimenting with automatically issuing a pull request against this repo in our release process. It looks like we need to actually to push the branch to DataBiosphere (not our fork) before making the PR. I'll fix this in our release process.

Qi77Qi:
sounds great..thanks @danking ! btw https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-hail/Dockerfile#L7 needs to be updated as well

danking:
@Qi77Qi , it appears my user account does not have permission to make a PR into databiosphere/terra-docker. Can you grant me permission?

danking:
This is the bot: https://github.com/hail-tester1

Qi77Qi:
@danking added both...think both invites are waiting for confirmation

PR #309, opened 2022-03-18
update to hail 0.2.91

danking:


danking:
@Qi77Qi sorry this took so long! We tried to squeeze in a change before the release. Never a good idea 😉 . The PR looks right to me. The version is updated in the Dockerfile, there's a change log, and the image patch version was bumped. I verified the changelog URL we generate works too.

danking:
Hopefully, from here on out, you'll get a steady stream of automatic Hail version bumps!

Qi77Qi (approved):
lgtm! thanks @danking !

Just fyi, we do have another image also installs hail [here](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-aou/Dockerfile#L37). This image is used by AoU site, if your CI can update that image as well, that'd be awesome...but no worries if not :)

also cc @calbach

PR #316, opened 2022-04-27
[IA-3327] Added Nextflow to GATK

cstrandw:
Removed Nextflow installation from AoU image and put it into the GATK image

Tested by building and running a container locally and going through the Nextflow tutorial (https://www.nextflow.io/docs/latest/getstarted.html) to verify that Nextflow runs as expected. Saw Nextflow Hello World work correctly.

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.10
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.1.0
 
 USER root
 
@@ -180,9 +180,8 @@ RUN mkdir -p /tmp/vcftools && \
 # important here, as packages.install() will target the first path in this list,
 # and the jupyter user will have readonly access to the conda path.
 ENV R_LIBS /usr/local/lib/R/site-library:/opt/conda/lib/R/library
-RUN conda update -n base conda && \
-    conda install -c bioconda "r-base>=4.2" r-saige
+RUN conda install -c bioconda r-saige
 
 USER $USER
 
-#RUN R -e 'BiocManager::install(c("GENESIS"))'
+RUN R -e 'BiocManager::install(c("GENESIS"))'
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.1.0
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.10
 
 USER root
 
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.9
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.1.0
 
 USER root
 
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -180,8 +180,9 @@ RUN mkdir -p /tmp/vcftools && \
 # important here, as packages.install() will target the first path in this list,
 # and the jupyter user will have readonly access to the conda path.
 ENV R_LIBS /usr/local/lib/R/site-library:/opt/conda/lib/R/library
-RUN conda install -c bioconda r-saige
+RUN conda update -n base conda && \
+    conda install -c bioconda "r-base>=4.2" r-saige
 
 USER $USER
 
-RUN R -e 'BiocManager::install(c("GENESIS"))'
+#RUN R -e 'BiocManager::install(c("GENESIS"))'
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -173,17 +173,6 @@ RUN mkdir -p /tmp/vcftools && \
     cd $HOME && \
     rm -rf /tmp/vcftools
 
-# Install Nextflow
-ENV NXF_MODE google
-RUN mkdir -p /tmp/nextflow && \
-  cd /tmp/nextflow && \
-  curl -s https://get.nextflow.io | bash && \
-  mv nextflow /bin/nextflow && \
-  cd $HOME && \
-  chmod 777 /bin/nextflow && \
-  chown -R $USER:users $HOME/.nextflow && \
-  rm -rf /tmp/nextflow
-
 # TODO(IA-3261): Remove this hack, and just conda install normally. R SAIGE
 # unfortunately is difficult to install outside of conda. Conda is also not
 # fully supported on the terra-docker base image. Here we workaround this by
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -29,6 +29,8 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
 ENV PIP_USER=false
 ENV PYTHONPATH $PYTHONPATH:/usr/lib/spark/python
 
+RUN pip3 install --upgrade markupsafe==2.0.1
+
 RUN pip3 install \
       igv-jupyter \
       nbstripout \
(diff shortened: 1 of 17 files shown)


Qi77Qi (approved):
🎉

calbach (commented):
AoU change looks good to me. What's the criteria for bumping the minor version rather than the patch? Feels like a relatively minor change to me (especially relatively to other changes I've seen here, which only bump the patch number).

cstrandw:
> AoU change looks good to me. What's the criteria for bumping the minor version rather than the patch? Feels like a relatively minor change to me (especially relative to other changes I've seen here, which only bump the patch number).

We usually use patch version bumps for bug fixes. Because we are moving where we install Nextflow, we decided to go with a minor bump for both images affected since their tools (adding and removing Nextflow) changed.

Qi77Qi:
I don't feel strongly..but this PR doesn't really remove any package, it just adds a library...so patch would've been fine too

calbach (approved):
OK. Could we do 2.0.15 for AoU instead? From our image's perspective, this change is effectively a no-op.

cstrandw:
> OK. Could we do 2.0.15 for AoU instead? From our image's perspective, this change is effectively a no-op.

Yes, just pushed that change. Thanks!

calbach:
Per slack, doesn't look related to your changes, but looks like a legitimate issue that would likely result in a failure build the docker image if we merged.

```
W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
```

This is not a repo that we added ourselves, it was added in a base image. It's curious that we're not seeing issues building those images, perhaps it's due to a particular package we're installing, or perhaps we are and they just aren't being tested here.

cstrandw:
> Per slack, doesn't look related to your changes, but looks like a legitimate issue that would likely result in a failure build the docker image if we merged.
> 
> ```
> W: GPG error: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY A4B469963BF863CC
> ```
> 
> This is not a repo that we added ourselves, it was added in a base image. It's curious that we're not seeing issues building those images, perhaps it's due to a particular package we're installing, or perhaps we are and they just aren't being tested here.

Got it, that makes sense. Do you have any suggestions about how we should proceed with this?

terra-jupyter-r/Dockerfile
```diff
@@ -121,7 +121,7 @@ RUN pip3 -V \
 
 RUN R -e 'install.packages("BiocManager")' \
     ## check version
-    && R -e 'BiocManager::install(version="3.14", ask=FALSE)' \
+    && R -e 'BiocManager::install(version="3.15", ask=FALSE)' \
```
Qi77Qi, line 124: we're updating r version in this PR too?

gpcarr: bioconductor released 3.15 which got pulled in automatically to `terra-jupyter-r` - this line just checks the version of bioconductor that gets installed

Qi77Qi: why are we updating bioconductor in this PR?

gpcarr: we aren't explicitly updating the version - without any changes to the r image, [this line ](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-r/Dockerfile#L124)fails because we're installing bioconductor version 3.15

alternatively, we could try to pin version 3.14 in [this line](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-r/Dockerfile#L122) to avoid bumping the bioconductor version

Qi77Qi: ah yea...pinning that version is probably better... @nturaga has a seaparate PR for uprading bioconductor

gpcarr: @nturaga's [PR](https://github.com/DataBiosphere/terra-docker/pull/318) only updates Bioconductor in the Anvil RStudio image and not in the terra-jupyter-r image

Qi77Qi: he hasn't PRed that one yet...I paired with him the other day

nturaga: Yes, this is because the R-4.2.0 update on the apt repository isn't available yet. We need to wait to do this with `apt`. The other way to do this is to do a source installation of R, which would involve compiling everything from scratch.

Qi77Qi: let's separate things...and pin the version in https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-r/Dockerfile#L122 for this one

gpcarr: OH that makes SO much sense. I updated bioconductor for terra-jupyter-r and that was causing issues so I'll go ahead and try to pin version 3.14 until we're ready to bump R

gpcarr:
> Hi @calbach @yonghaoy, the AoU image is now failing due to an R and Bioconductor version update breaking the Conda installation. The image build is [failing on this line](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-aou/Dockerfile#L198) with this [error message](https://gist.github.com/cstrandw/3b320ab3ab2877ef402d24426e7d2f8a), because the [installation of Conda on this line](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-aou/Dockerfile#L194) causes the R version to be reverted to 4.1 when version 4.2 is expected by the new version of bioconductor (3.15).
> 
> @gpcarr spent a couple of days trying to fix this issue through Conda and was unable to find a good solution at this time, so we believe a potential path forward would be to revert all of the changes to the AoU image and create a follow up ticket to resolve this in the future since we do not support Conda. Potential fixes would be fixing the Conda installation or reverting bioconductor to 3.14. Does this sound like a good plan to you, or do you have any other potential solutions to this issue?

@calbach just fyi, we ended up not including the bioconductor and r version bump in this PR so this comment will no longer be relevant here but will come up in the [PR to bump bioconductor and r](https://github.com/DataBiosphere/terra-docker/pull/319) which we are looking to merge after this PR gets merged.

gpcarr:
bioconductor test failed - https://github.com/DataBiosphere/terra-docker/runs/6262730955?check_suite_focus=true

looks like we are trying to install a package called `RMySQL` which requires that we install `libmysql` in the docker image (likely the r image) which we are not currently installing

will leave it up to the bioconductor team if they'd prefer removing the test or adding that package @nturaga maybe you can include that in your [PR](https://github.com/DataBiosphere/terra-docker/pull/319)

@cstrandw we'll be good to merge once aou and gatk-ovtf tests pass!

terra-jupyter-base/Dockerfile
```diff
@@ -100,6 +98,7 @@ RUN pip3 -V \
  && pip3 install python-datauri \
  && pip3 install jupyter_contrib_nbextensions \
  && pip3 install jupyter_nbextensions_configurator \
+ && pip3 install markupsafe==2.0.1 \
```
gpcarr, line 101: This was needed in every python image starting with the base. MarkupSafe released a new version that is not compatible with a package that `samtools` installs. We reinstall samtools or update packages in every image which was why we needed to add this line to all python images.

terra-jupyter-r/Dockerfile
```diff
@@ -38,8 +38,8 @@ RUN apt-get update \
 	libgit2-dev \
 	libxml2-dev \
 	libgfortran-7-dev \
-	r-base-dev \
-	r-base-core \
+	r-base-dev=4.1.3-1.1804.0 \
+	r-base-core=4.1.3-1.1804.0 \
```
gpcarr, line 42: we had to pin the R version to 4.1 until bioconductor is ready to release bioconductor 3.15 with r 4.2

terra-jupyter-aou/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-gatk:2.0.9
```
Qi77Qi, line 1: this is a minor version bump...is it intended?

gpcarr: yes the minor bump is for gatk only. we bumped aou as a patch https://github.com/DataBiosphere/terra-docker/pull/316/files#diff-111700937e15194bfcc732196ff2c66042b8496f9077dc25383577a79e174092R1

we were always going to bump gatk as a minor bump because nextflow is a big/involved tool that we are now offering in gatk which we did not previously offer in that image

PR #325, opened 2022-05-31
[IA-3261] Users are getting permissions errors when running conda install.

n8landolt:
## Description

User runs conda install <package> and gets permissions error.
The conda installation and associated packages are owned by root:root, so when the "jupyter" user goes to install they will not have permissions to make changes.

## Solution
Create a virtual environment for the user to allow them to install packages in their home directory.
Implementation: clone the base environment into a new environment in the user's home directory. That way they have all of the packages expected, and they can additionally install their own.


n8landolt:
Initially I had the chown only in the base image, however, the extended Dockerfiles are installing other conda packages as the root user. 

Additionally, do we want any of our images that extends terra-jupyter-base to be able to install conda packages? We may need to run chown on all dockerfiles.

terra-jupyter-base/Dockerfile
```diff
@@ -21,7 +21,6 @@ RUN sudo -i \
 # see article: https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/
 # first we remove the bad keys from base image (https://github.com/NVIDIA/nvidia-docker/issues/1631#issuecomment-1112828208)
 RUN rm /etc/apt/sources.list.d/cuda.list \
-    && rm /etc/apt/sources.list.d/nvidia-ml.list \
```
n8landolt, line 24: In the 2-9 version of deeplearning base image, this file does not exist and therefore cannot be removed.

terra-jupyter-base/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # adapted from https://hub.docker.com/r/jupyter/base-notebook/ AKA https://github.com/jupyter/docker-stacks/tree/master/base-notebook
 
-FROM gcr.io/deeplearning-platform-release/tf-gpu.2-7
+FROM gcr.io/deeplearning-platform-release/tf-gpu.2-9
```
n8landolt, line 3: The updated image contains Conda 4.12.0.

The old image contained Conda 4.11.0 which was causing issues.

PR #326, opened 2022-06-02
[IA-3042] update contributing guide

xuscrbw:
Added some new best practices with the terra-docker-versions.json. So added a line in the contributing guide to mention that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,7 +75,7 @@ The terra-docker image hash update [job](https://fc-jenkins.dsp-techops.broadins
 
 ## Updating terra-docker-versions-*.json Best Practices
 
-As of 6/8/2022, we have added a new process for updating the image versions available across different environments. Once the Leo PR is merged into develop, the user will need to generate the `terra-docker-versions-new.json` with the new images and copy it by doing `gsutil cp xxx/terra-docker-versions-new.json terra-docker-image-documentation/terra-docker-versions-candidate.json`. The remainder of the update process will be covered through Leo deployment. As part of reliability changes, the versions file will live in the corresponding bucket with the following name convention: `terra-docker-image-documentation-{environment}`.
+As of 6/8/2022, we have added a new process for updating the image versions available across different environments. Before the Leo PR is merged into develop, the user will need to generate the `terra-docker-versions-new.json` with the new images and copy it by doing `gsutil cp xxx/terra-docker-versions-new.json terra-docker-image-documentation/terra-docker-versions-candidate.json`. The remainder of the update process will be covered through Leo deployment. As part of reliability changes, the versions file will live in the corresponding bucket with the following name convention: `terra-docker-image-documentation-{environment}`.
 
 ## Merging the terra-docker image hash update job PR
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,10 @@ There will be a notification in the slack channel `#dsp-callisto-internal` once 
 
 The terra-docker image hash update [job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/)  mentioned above determines which images (including derived images) have been changed and builds each image with the appropriate tag.
 
+## Updating terra-docker-versions-*.json Best Practices
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -73,6 +73,10 @@ There will be a notification in the slack channel `#dsp-callisto-internal` once
 
 The terra-docker image hash update [job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/)  mentioned above determines which images (including derived images) have been changed and builds each image with the appropriate tag.
 
+## Updating terra-docker-versions-*.json Best Practices
+
+As of 6/2/2022, we have added a new process for updating the image versions available across different environments. Once the Leo PR is merged to develop, we will need to bump the `terra-docker-versions-dev.json` file with the new version file with updated entries for each new image. Once the dev candidate is promoted to alpha, we will need to bump the subsequent `terra-docker-versions-alpha.json` file. This process should be repeated for promotions to staging with `terra-docker-versions-staging.json` and then to prod with `terra-docker-versions-prod.json`. This will mainly impact our terra-ui integration tests against alpha and staging, but if we forget, we should be running into errors when creating runtimes in the environment which has an outdated versions file.
```
Qi77Qi, line 78: I think it's more like
```
...Once the Leo PR is merged to develop, we will need to promote `terra-docker-versions-new.json` to dev by doing `gsutil cp xxx/terra-docker-versions-new.json xx/terra-docker-versions-dev.json.
```

For the other environments, let's rely on automation instead of manual update

CONTRIBUTING.md
```diff
@@ -73,6 +73,10 @@ There will be a notification in the slack channel `#dsp-callisto-internal` once
 
 The terra-docker image hash update [job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/)  mentioned above determines which images (including derived images) have been changed and builds each image with the appropriate tag.
 
+## Updating terra-docker-versions-*.json Best Practices
+
+As of 6/8/2022, we have added a new process for updating the image versions available across different environments. Once the Leo PR is merged into develop, the user will need to generate the `terra-docker-versions-new.json` with the new images and copy it by doing `gsutil cp xxx/terra-docker-versions-new.json terra-docker-image-documentation/terra-docker-versions-candidate.json`. The remainder of the update process will be covered through Leo deployment. As part of reliability changes, the versions file will live in the corresponding bucket with the following name convention: `terra-docker-image-documentation-{environment}`.
```
Qi77Qi, line 78: can you have the full path here instead `xxx`?

xuscrbw: The path here would be where the user has the -new.json file. I'm not sure where the file would actually get generated in this case. I think the custom image job should generate the file into the terra-docker-image-documentation bucket, but I don't think it's worked for a little. I can update it in any case

Qi77Qi: we should update it to what the CI generates it to...even if ppl are doing it manually when CI doesn't work, we should generate it in the right place instead of random places

CONTRIBUTING.md
```diff
@@ -73,6 +73,10 @@ There will be a notification in the slack channel `#dsp-callisto-internal` once
 
 The terra-docker image hash update [job](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/)  mentioned above determines which images (including derived images) have been changed and builds each image with the appropriate tag.
 
+## Updating terra-docker-versions-*.json Best Practices
+
+As of 6/8/2022, we have added a new process for updating the image versions available across different environments. Before the Leo PR is merged into develop, the user will need to generate the `terra-docker-versions-new.json` with the new images and copy it by doing `gsutil cp terra-docker-image-documentation/terra-docker-versions-new.json terra-docker-image-documentation/terra-docker-versions-candidate.json`. The remainder of the update process will be covered through Leo deployment. As part of reliability changes, the versions file will live in the corresponding bucket with the following name convention: `terra-docker-image-documentation-{environment}`.
```
Qi77Qi, line 78: https://github.com/DataBiosphere/terra-docker/pull/326#discussion_r891529079 and it should be `gsutil cp gs://..`

Qi77Qi (approved):

PR #328, opened 2022-06-03
[IA-3261] Added chown command for /opt/conda.

n8landolt:
## Description

User runs conda install <package> and gets permissions error.
The conda installation and associated packages are owned by root:root, so when the "jupyter" user goes to install they will not have permissions to make changes.

## Solution
Enable users to have greater control over Conda installs by changing ownership of /opt/conda

Also update to newer version of conda by upgrading the deeplearning base image.

### DIFF
--- a/terra-jupyter-bioconductor/Dockerfile
+++ b/terra-jupyter-bioconductor/Dockerfile
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:2.1.3
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:2.1.4
 
 USER root
 
@@ -6,5 +6,7 @@ USER root
 RUN curl -O https://raw.githubusercontent.com/Bioconductor/anvil-docker/master/anvil-rstudio-bioconductor/install.R \
 	&& R -f install.R \
 	&& rm -rf install.R
+
+RUN chown -R $USER:users /opt/conda
 
 USER $USER
--- a/config/conf.json
+++ b/config/conf.json
@@ -26,7 +26,7 @@
                     "tidyverse"
                 ]
             },
-            "version" : "2.1.3",
+            "version" : "2.1.4",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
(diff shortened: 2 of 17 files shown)


terra-jupyter-bioconductor/Dockerfile
```diff
@@ -7,4 +7,6 @@ RUN curl -O https://raw.githubusercontent.com/Bioconductor/anvil-docker/master/a
 	&& R -f install.R \
 	&& rm -rf install.R
 
+RUN chown -R $USER:users /opt/conda
```
Qi77Qi, line 10: does this need to be in each individual image even tho it's already in base image?

Qi77Qi, line 10: does this need to be in each individual image even tho it's already in base image?

n8landolt: I'm uncertain, but I think it's safer this way. There are additional packages that get installed as the root user throughout the Dockerfiles.

Qi77Qi: are you saying `pip` install will install packages to `/opt/conda`?

n8landolt: Yes, if you run `conda list` It will show the source as pip instead of conda-forge

Qi77Qi: can you cd into `/opt/conda` to confirm? just curious if the packages are indeed in this dir, or it's just `conda` is looking at `pip` package location as well

n8landolt: Pip is also installed under /opt/conda/bin/pip, as well as the packages being located under /opt/conda/lib/python3.7/site-packages/

Requests is under that directory, and that is installed via pip3 in the dockerfile.

n8landolt: I can test if it's necessary to run the chown in the latter Dockerfiles if you would like. I'm uncertain if it's necessary, it'll just take me a while to test.

Qi77Qi (approved):

Qi77Qi:
tests still need to pass before merging

calbach:
Did the tests ever pass here? I'm now seeing failures on my PR which may be related to this:


```
Step 10/14 : RUN chown -R $USER:users /opt/conda
 ---> Running in 62548cd22f84
ApplyLayer exit status 1 stdout:  stderr: write /opt/conda/lib/libmkl_core.so.2: no space left on device
Error: Process completed with exit code 1.
```


https://github.com/DataBiosphere/terra-docker/runs/6819080486?check_suite_focus=true

PR #330, opened 2022-06-08
[IA-3401] Fix broken Python packages in gatk / AoU images

calbach:
See https://broadworkbench.atlassian.net/jira/software/c/projects/IA/issues/IA-3401 for a description of the bug.

I narrowed the cause down to the GATK image (the issue did not repro in -base, -r, -python). I then further narrowed it down to the way in which sites-packages were being copied from the python image. I'm not 100% sure of why the previous approach caused this bug, but my suspicion is that bringing packages through the backdoor like this missed some initialization or setup with the base conda environment; the bug seems related to installing with conda vs pip.

By switching to a requirements.txt file in terra-jupyter-python, we can reuse the dependencies in both images via `pip install -r`.

This also:
- removes the setuptools downgrade in the python image, pysam now installs, so this seems to no longer be necessary
- bumps hail to 0.2.93
- removes the java8 logic  for Hail (this seems to have not made a difference...)

Known issues:
- `cloud alpha storage cp` is not working due to the crc32c package not being fully installed. I've tried a few fixes but failed - leaving that for now.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -68,7 +68,7 @@
                     "scikit-learn"
                 ]
             },
-            "version" : "1.0.9",
+            "version" : "1.0.11",
             "automated_flags" : {
                 "generate_docs" : true,
                 "include_in_ui" : true,
@@ -124,7 +124,7 @@
             "packages" : {
                 
             },
-            "version" : "2.2.3",
+            "version" : "2.2.5",
             "automated_flags" : {
                 "include_in_ui" : true,
                 "generate_docs" : true,
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "2.1.3",
+            "version" : "2.1.5",
             "automated_flags" : {
                 "include_in_ui" : false,
                 "generate_docs" : false,
--- a/terra-jupyter-aou/CHANGELOG.md
+++ b/terra-jupyter-aou/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.1.5 - 2022-06-09
 
-- Update `terra-jupyter-gatk` base to `2.2.4`
+- Update `terra-jupyter-gatk` base to `2.2.5`
   - Fix Python sites-packages to avoid issues on uninstall
 - Bump Hail to 0.2.93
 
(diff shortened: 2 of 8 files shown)


calbach:
Looks like this approach misses pandas_gbq and others... investigating..

calbach:
Please hold on review, need to switch the approaches here... I have an alternative that seems to work - verifying locally now.

calbach:
OK - PTAL

calbach:
```
Step 10/14 : RUN chown -R $USER:users /opt/conda
 ---> Running in 218cb2c9e97c
ApplyLayer exit status 1 stdout:  stderr: write /opt/conda/lib/libmkl_core.so.2: no space left on device
```

config/conf.json
```diff
@@ -68,7 +68,7 @@
                     "scikit-learn"
                 ]
             },
-            "version" : "1.0.9",
+            "version" : "1.0.11",
```
calbach, line 71: I'm going to skip `1.0.10`, since I don't know if reusing that ID is going to cause issues (I'm guessing some of these built behind the scenes)

Qi77Qi (approved):
once tests pass 👍

config/conf.json
Qi77Qi: I could also just remove these bad tags in GCP...lmk

calbach: Let me know if you prefer that - my default would be to keep the PR as-is so I don't need to rerun the tests again (and avoiding reuse seems slightly safer).

Qi77Qi: @n8landolt do you have a preference? basically in your new PR, if we don't remove the old tags in GCP, then you'll need to bump all image versions again...

config/conf.json
```diff
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "2.1.3",
+            "version" : "2.1.5",
```
Qi77Qi, line 146: I just checked... there was no 2.1.4 generated

config/conf.json
n8landolt: I think we want to just bump it to .11 No real harm in missing a minor version, right?

PR #334, opened 2022-06-15
Fix the Hail Dockerfile

danking:
We started using pip system specifiers (e.g. ; sys_platform != 'win32') which was causing issues with xargs. Writing a file and running on the file seems simpler anyway.

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -63,10 +63,12 @@ RUN find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
     && pip3 install pypandoc gnomad \
     && pip3 install --no-dependencies hail==$HAIL_VERSION \
     && X=$(mktemp -d) \
+    && requirements_file=$(mktemp) \
     && mkdir -p $X \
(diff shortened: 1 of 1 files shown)


calbach (approved):
Thanks. Please check my understanding: the prior approach was not yet a problem, but would be on a future version of Hail.

danking:
Yeah that's exactly right @calbach. 0.2.94 includes the sys_platform change.

danking:
Hmm, seems like the docker build inadvertently triggered an update of some part of Jupyter:
```
KeyError: 'The Python module igv is not a valid nbextension, it is missing the `_jupyter_nbextension_paths()` method.'
```

calbach:
Trying that fix here: https://github.com/DataBiosphere/terra-docker/pull/335

calbach:
Merged #335  with fix, hopefully this PR should be good on rebase.

danking:
@calbach fwiw, I think checks are not required on this repo right now (I still see the squash and merge button).

calbach:
I don't know if that's intended or not, I merely have some authority over this image in particular, not the overall repo. @Qi77Qi FYI

Qi77Qi (approved):

PR #337, opened 2022-06-23
[RW-8461] Bump base image and all sub images to fix leo_url in cromwell python script

yonghaoy:
The change that I need: https://github.com/broadinstitute/cromwhelm/pull/34

### DIFF
--- a/terra-jupyter-aou/CHANGELOG.md
+++ b/terra-jupyter-aou/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 2.1.6 - 2022-06-23T10:58:12.961300Z
 
 - Update `terra-jupyter-base` to `1.0.10`
-  - Fix leo_url variable in workspace_cromwell.py ****script for AoU projects
+  - Fix leo_url variable in workspace_cromwell.py script for AoU projects
 
 ## 2.1.5 - 2022-06-09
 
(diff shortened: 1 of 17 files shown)


terra-jupyter-aou/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.1.6 - 2022-06-23T10:58:12.961300Z
+
+- Update `terra-jupyter-base` to `1.0.10`
+  - Fix leo_url variable in workspace_cromwell.py ****script for AoU projects
```
calbach, line 4: what's with the `****` ?

calbach (approved):
FYI: as for why this change does anything, it's because this line pulls from master: https://github.com/DataBiosphere/terra-docker/blob/c45d2d22075990ebbdd7373606bf8dc47de650a5/terra-jupyter-base/Dockerfile#L142

Qi77Qi:
can we use a permlink for the script? that way we'll know exactly what version of the script we're pulling

calbach:
From offline debugging with @yonghaoy , it seems likely that at least some of these failures (e.g. out of disk) are caused by conflicts with existing versions from this reverted PR. @Qi77Qi I think we should flag any version created in this PR as "do not use" in respective changelogs, if there are any others that Yonghao is not addressing in the PR here.

Qi77Qi (approved):

PR #342, opened 2022-07-08
[IA-3261] Updated dockerfiles and changelogs for conda install.

n8landolt:
## Description

Users need to be able to run Conda install in their environment in AoU.

## Solution

From within the dockerfile, we create an environment where the user has write access. 

If the user wants to install conda packages, they must select the appropriate environment kernel. (The base image env is user-base, the aou image env is user-aou).

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -26,7 +26,7 @@
                     "tidyverse"
                 ]
             },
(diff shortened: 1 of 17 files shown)


Qi77Qi:
seems test for aou is skipped..is that intentional?

PR #345, opened 2022-07-19
[SUP-496] Add AnVIL to r Dockerfile

n8landolt:
https://broadworkbench.atlassian.net/browse/SUP-496

## Description
AnVIL is an essential tool for R to interface with TERRA tables and workspaces.
We need to install and verify the installation for the r image.


terra-jupyter-bioconductor/Dockerfile
```diff
@@ -1,10 +1 @@
...
-
-USER root
-
-## Install Bioconductor Core packages
-RUN curl -O https://raw.githubusercontent.com/anvilproject/anvil-docker/master/anvil-rstudio-bioconductor/install.R \
-	&& R -f install.R \
-	&& rm -rf install.R
-
-USER $USER
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:2.1.6
```
Qi77Qi, line 1: @nturaga would like to run these changes by you

nturaga: This looks good to me. 

It seems the changes that were made are moving the contents of the `install.R` file into the Dockerfile explicitly to make the `terra-jupyter-r` image the same as the `terra-jupyter-bioconductor` image. 

Correct me if i'm wrong about this.

nturaga (approved):

Qi77Qi (approved):

terra-jupyter-bioconductor/Dockerfile
n8landolt: You are correct. Part of the story is we wanted the AnVIL packages in the R image, but it just to happened to make the bioconductor image identical to the terra-jupyter-r image.

Issue #333, opened 2022-06-15
Shrink size of terra-jupyter-base


SHuang-Broad:
Currently, it's 8.4GB

https://console.cloud.google.com/gcr/images/broad-dsp-gcr-public/US/terra-jupyter-base@sha256:8b8da2a3ac90e04015694b0fae20518eb38db4a5f7bc18144d3fbf81e8d27066/details?tag=latest

Pulling that takes a long time (on my config, it took over 25 minutes, though I expect it to be faster on GCP's network). This eats into the 30 minutes limit of creating one's custom environment on Terra.

Thank you.

Qi77Qi:
thanks for reporting...The image is big becuz we're using https://cloud.google.com/deep-learning-containers/docs/choosing-container and these images are around 10G themselves.

We do want to somehow make things better, but it's not currently prioritized yet..

SHuang-Broad:
To follow up, I just ran an experiment using `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:latest` as the custom image, and the following startup script

```bash
#!/bin/bash

set -eu

echo "test"
```

It timed out.

Surely this is just one data point. But it's been consistent with my previous experiences with other startup scripts as well, where the startup script was just an installation of samtools and bcftools, and it took close to 30 minutes to finish.

We could build a custom image based on `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:latest`, but it will be inevitably bigger and may time out easily, given this large initial overhead.


Can you please help on this?

SHuang-Broad:
I see.

It's imaginable some of us not needing the DL functionalities, but do need some of the Jupyter-related stuff, i.e. what @sjfleming has suggested.

So I think his suggestion makes sense, i.e. to truly make this a "base" image, as from IMHO my base case doesn't need GPU/DL.

That being said, I understand it's not a priority for you. So we'd appreciate it if there are instructions for a PR to make that happen.

sjfleming:
Just to add another observation, I see that `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r` is currently 20.1 GB

sjfleming:
And @Qi77Qi , as @SHuang-Broad mentioned, I would also be happy to contribute to a PR to make this happen, if we know what would be acceptable in terms of requirements

Qi77Qi:
[...]

PR #346, opened 2022-08-05
Discussion: draft of a minimal "developer" base image

sjfleming:
This is not a real PR, which I'm sure would require a lot more detail work.  This is a draft so that we can discuss the general strategy toward resolving #333 .

Some preliminary discussion (design doc) is here
https://docs.google.com/document/d/1b9uXanA3uCxUJoKktczxFXpJ1o3TYeXnaM0PGvbAczM/edit#

------------------------

__What has been done:__
- This adds a folder called `./terra-jupyter-dev-base` to the repo that contains two files:
    - `Dockerfile`
    - `build_docker.sh`: you can run this script to build the image, which is called `terra-jupyter-dev-base:0.0.1`

__Results so far:__
- A (seemingly) working image built on `ubuntu:20.04` that can successfully run a Jupyter notebook server using the command `docker run --rm -it -p 8888:8000 terra-jupyter-dev-base:0.0.1`

__Testing:__
- Very little...
    - No integration testing to see if this actually works in Terra
    - No testing to see if the scripts function (the scripts that do welder-related things? and the custom Jupyter extensions?)
- If I run `docker run --rm -it -p 8888:8000 terra-jupyter-dev-base:0.0.1`, then I can 
    - Successfully connect to the notebook server from a browser
    - See my python3 kernel, which points to the correct python3 install in the image
    - Successfully run simple commands in the Jupyter notebook, i.e. the kernel works
- No actual tests have been added to the repo: I don't understand enough about the repo to do this
    - But tests should be exactly the same as for `terra-jupyter-base`?

### DIFF
--- a/terra-jupyter-dev-base/Dockerfile
+++ b/terra-jupyter-dev-base/Dockerfile
@@ -19,39 +19,41 @@ RUN apt-get update && apt-get install -yq --no-install-recommends \
     # gnupg requirement
     dirmngr \
     gnupg \
-    # curl requirement
+#    # curl requirement
     curl \
     ca-certificates \
     # useful utilities for debugging within the docker
     nano \
     procps \
-    lsb-release \
-    # python requirements
-    checkinstall \
-    build-essential \
-    zlib1g-dev \
-    # pip requirements
-    libssl-dev \
-    libbz2-dev \
-    libreadline-dev \
-    libsqlite3-dev \
-    llvm \
-    libncurses5-dev \
-    libncursesw5-dev \
-    tk-dev \
-    libffi-dev \
-    liblzma-dev \
-    python-openssl \
-    libexempi3 \
-    libv8-3.14-dev \
+#    lsb-release \
+#    # python requirements
+#    checkinstall \
+#    build-essential \
+#    zlib1g-dev \
+#    # pip requirements
+#    libssl-dev \
+#    libbz2-dev \
+#    libreadline-dev \
+#    libsqlite3-dev \
+#    llvm \
+#    libncurses5-dev \
+#    libncursesw5-dev \
+#    tk-dev \
+#    libffi-dev \
+#    liblzma-dev \
+#    python-openssl \
+#    libexempi3 \
+#    libv8-3.14-dev \
+    # extras \
+    wget \
+    bzip2 \
     # install script requirements
-    sudo \
     locales \
     # for ssh-agent and ssh-add
     keychain \
-    # openjdk 11
-    default-jre \
-    default-jdk \
+##    # openjdk 11
+##    default-jre \
+##    default-jdk \
     # git
     git \
 # Uncomment en_US.UTF-8 for inclusion in generation
@@ -86,16 +88,58 @@ ENV JUPYTER_HOME /etc/jupyter
 
 # install miniconda to /opt/conda
 ENV CONDA_AUTO_UPDATE_CONDA=false
-ENV PATH="${PATH}:/opt/conda/bin:${HOME}/.local/bin:${HOME}/packages/bin"
+ENV CONDA_DIR /opt/conda
+ENV PATH="${PATH}:${CONDA_DIR}/bin:${HOME}/.local/bin:${HOME}/packages/bin"
 RUN curl -so $HOME/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-py37_4.12.0-Linux-x86_64.sh \
  && chmod +x $HOME/miniconda.sh \
- && $HOME/miniconda.sh -b -p /opt/conda \
+ && $HOME/miniconda.sh -b -p $CONDA_DIR \
  && rm $HOME/miniconda.sh
+
+# slim install of things that have a gcc dependency
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    checkinstall \
+    build-essential \
+    zlib1g-dev \
+    # pip requirements
+    libssl-dev \
+    libbz2-dev \
+    libreadline-dev \
+    libsqlite3-dev \
+    llvm \
+    libncurses5-dev \
+    libncursesw5-dev \
+    tk-dev \
+    libffi-dev \
+    liblzma-dev \
+    python-openssl \
+    libexempi3 \
+    libv8-3.14-dev \
+ && rm -rf /var/lib/apt/lists/* \
+ && pip3 install firecloud  \
+ && pip3 install terra-notebook-utils \
+ && apt-get purge -y --auto-remove \
+    checkinstall \
+    build-essential \
+    zlib1g-dev \
+    # pip requirements
+    libssl-dev \
+    libbz2-dev \
+    libreadline-dev \
+    libsqlite3-dev \
+    llvm \
+    libncurses5-dev \
+    libncursesw5-dev \
+    tk-dev \
+    libffi-dev \
+    liblzma-dev \
+    python-openssl \
+    libexempi3 \
+    libv8-3.14-dev
 
 RUN pip3 -V \
  # For gcloud alpha storage support.
  && pip3 install google-crc32c --target /usr/lib/google-cloud-sdk/lib/third_party \
- && git clone --branch sf_missing_requirement --single-branch \
+ && git clone --branch cromshell_2.0 --single-branch \
     https://github.com/broadinstitute/cromshell.git $HOME/cromshell \
  # tmp hack min-5
  # I'm not installing jupyterlab and I can't update init-actions.sh to not access it
@@ -104,7 +148,10 @@ RUN pip3 -V \
  # Hence, make sure to manually test out "launch terminal" button (the button in the green bar next to start and stop buttons)
  # to make sure we don't accidentally break it every time we upgrade notebook version until we figure out an automation test for this
  && pip3 install notebook \
- && pip3 install ipykernel \
+ && pip3 install jupyter \
+# && pip3 install jupyterlab \
+ && pip3 install cookiecutter \
+ && pip3 install tornado \
  && pip3 install python-datauri \
  && pip3 install jupyter_contrib_nbextensions \
  && pip3 install jupyter_nbextensions_configurator \
@@ -113,8 +160,8 @@ RUN pip3 -V \
  && pip3 install "nbconvert>=6.4.5" \
  # for jupyter_delocalize.py and jupyter_notebook_config.py
  && pip3 install requests \
- && pip3 install firecloud \
- && pip3 install terra-notebook-utils \
+# && pip3 install firecloud \
+# && pip3 install terra-notebook-utils \
  && pip3 install crcmod \
  # install Cromshell 2.0, using repo cloned above
  && pip install $HOME/cromshell
@@ -144,7 +191,7 @@ RUN chown -R $USER:users $JUPYTER_HOME \
 # && jupyter nbextension disable nb_conda --py --sys-prefix \
  && find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
 #  You can get kernel directory by running `jupyter kernelspec list`
- && $JUPYTER_HOME/scripts/kernel/kernelspec.sh $JUPYTER_HOME/scripts/kernel /opt/conda/share/jupyter/kernels
+ && $JUPYTER_HOME/scripts/kernel/kernelspec.sh $JUPYTER_HOME/scripts/kernel $CONDA_DIR/share/jupyter/kernels
 
 USER $USER
 EXPOSE $JUPYTER_PORT
@@ -154,4 +201,4 @@ WORKDIR $HOME
 # When Leonardo deploys this image onto a cluster, the entrypoint is overwritten to enable
 # additional setup inside the container before execution.  Jupyter execution occurs when the
 # init-actions.sh script uses 'docker exec' to call run-jupyter.sh.
-ENTRYPOINT ["/opt/conda/bin/jupyter", "notebook"]
+ENTRYPOINT ["jupyter", "notebook"]
--- a/terra-jupyter-dev-base/Dockerfile
+++ b/terra-jupyter-dev-base/Dockerfile
@@ -2,48 +2,49 @@ FROM ubuntu:20.04
 
 USER root
 
-#######################
-# Prerequisites
-#######################
-ENV DEBIAN_FRONTEND noninteractive
+# The welder uid is consistent with the Welder docker definition here:
+# https://github.com/DataBiosphere/welder/blob/master/project/Settings.scala
+# Adding welder-user to the Jupyter container isn't strictly required, but it makes welder-added
+# files display nicer when viewed in a terminal.
+ENV DEBIAN_FRONTEND=noninteractive \
+    LC_ALL=en_US.UTF-8 \
+    USER=jupyter \
+    WELDER_USER=welder-user \
+    WELDER_UID=1001 \
+    # ensure this matches c.NotebookApp.port in jupyter_notebook_config.py
+    JUPYTER_PORT=8000 \
+    JUPYTER_HOME=/etc/jupyter \
+    CONDA_AUTO_UPDATE_CONDA=false \
+    CONDA_DIR=/opt/conda
+ENV HOME=/home/$USER
+    # When using PIP_USER=true packages are installed into Python site.USER_BASE, which is '/home/jupyter' for this system.
+    # Append '/home/jupyter/.local/bin' to PATH
+    # pip docs: https://pip.pypa.io/en/stable/reference/pip_install/#cmdoption-user
+ENV PATH="${PATH}:${CONDA_DIR}/bin:${HOME}/.local/bin:${HOME}/packages/bin"
 
-RUN apt-get update && apt-get install -yq --no-install-recommends sudo \
+# Users
+RUN useradd -m -s /bin/bash $USER \
+ && usermod -g users $USER \
+ && useradd -m -s /bin/bash -N -u $WELDER_UID $WELDER_USER \
+# Prerequisites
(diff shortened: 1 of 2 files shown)


sjfleming:
Anybody please feel free to jump in with any comments or advice
@Qi77Qi @SHuang-Broad

terra-jupyter-dev-base/Dockerfile
```diff
@@ -0,0 +1,163 @@
...
+
+# Users
+RUN useradd -m -s /bin/bash $USER \
+ && usermod -g users $USER \
+ && useradd -m -s /bin/bash -N -u $WELDER_UID $WELDER_USER \
+# Prerequisites
+ && apt-get update && apt-get install -yq --no-install-recommends \
+    sudo \
+ && sudo -i \
+    echo "deb http://security.ubuntu.com/ubuntu/ bionic main" >> /etc/apt/sources.list \
```
SHuang-Broad, line 33: `bionic` is code for 18.04. Is this supposed to be `Focal Fossa`?

terra-jupyter-dev-base/Dockerfile
```diff
@@ -0,0 +1,163 @@
...
+    tk-dev \
+    libffi-dev \
+    liblzma-dev \
+    python-openssl \
+    libexempi3 \
+    libv8-3.14-dev \
+# Install jupyter and some necessary python packages
+ && pip3 -V \
+ # For gcloud alpha storage support.
+ && pip3 install google-crc32c --target /usr/lib/google-cloud-sdk/lib/third_party \
```
SHuang-Broad, line 111: I'm a little concerned about mixing pip and conda, but I know this existed before this PR.

terra-jupyter-dev-base/Dockerfile
```diff
@@ -0,0 +1,163 @@
...
+ && git clone --branch cromshell_2.0 --single-branch \
+    https://github.com/broadinstitute/cromshell.git $HOME/cromshell \
+ # tmp hack min-5
+ # I'm not installing jupyterlab and I can't update init-actions.sh to not access it
+ && mkdir -p /usr/local/share/jupyter/lab \
+ # When we upgraded from jupyter 5.7.8 to 6.1.1, we broke terminal button on terra-ui.
+ # Hence, make sure to manually test out "launch terminal" button (the button in the green bar next to start and stop buttons)
+ # to make sure we don't accidentally break it every time we upgrade notebook version until we figure out an automation test for this
+ && pip3 install notebook \
```
SHuang-Broad, line 120: Do we want to pin down the versions? And/or have a `requirements.txt`?

Qi77Qi:
I thikn you need https://docs.google.com/document/d/1b9uXanA3uCxUJoKktczxFXpJ1o3TYeXnaM0PGvbAczM/edit?disco=AAAAdToq5Wg still (those scripts)

terra-jupyter-dev-base/Dockerfile
sjfleming: Good point: I am not sure about this.  I copied it directly from `terra-jupyter-base`. I'm honestly not sure what the base operating system actually is in `terra-jupyter-base`, since it uses a google deep learning base image `gcr.io/deeplearning-platform-release/tf-gpu.2-7`, and I can't seem to find the details.

The only thing I know is that, if I remove these steps from the build, it doesn't work.  These commands seem to be necessary for successful install of `libexempi3` and `libv8-3.14-dev` (which in turn are only needed in order to install the python packages `firecloud` and `terra-notebook-utils`)

terra-jupyter-dev-base/Dockerfile
jdcanas: requirements.txt would be a nice addition!

sjfleming: Probably makes sense to pin them

terra-jupyter-dev-base/Dockerfile
SHuang-Broad: Oh, I didn't mean to delete these. I mean for 20.04, shouldn't there be a new name other than `bionic`?

sjfleming:
@Qi77Qi is that something other than these?
https://github.com/sjfleming/terra-docker/blob/186a8b7225a04a2a704f26bbb939fdf4bd3e03b8/terra-jupyter-dev-base/Dockerfile#L140-L143

(I didn't see any other scripts being added to `terra-jupyter-base`)

terra-jupyter-dev-base/Dockerfile
sjfleming: Right, yeah, you could be right.  I'm not sure.  Somehow this works as is.  Is it maybe a workaround to install `bionic` packages in later ubuntu releases?  Or maybe it does need to be updated from `bionic` to `focal fossa` here.  I would need to ask the person who wrote this part :)

jdcanas:
You should be able to push the image to gcr or the like, and specify it as a custom image to test it in terra 

image

sjfleming:
@jdcanas Do you think I'm ready to try that out?  Is that the next step?

terra-jupyter-dev-base/Dockerfile
SHuang-Broad: I believe it's now `focal` replacing `bionic` for 20.04.

SHuang-Broad:
If what @sjfleming experimented above is allowed, then the last part of this section needs to be updated as well

https://github.com/databiosphere/terra-docker#terra-base-images

Screen Shot 2022-08-05 at 3 55 43 PM

sjfleming:
Okay so I figured, while I'm at it, let's make an equivalent base image that enables GPU functionality and has all the CUDA stuff pre-installed.  This might be out of scope for this current PR, but I'm going to include the Dockerfile here in case anyone wants to look at it.  The size of the image is 2.63 GB.

I have hosted it here for the time being, if anyone wants to play with it
```
us.gcr.io/broad-dsde-methods/terra-jupyter-gpu-dev-base:0.0.1
```

Tested:
- If I do `nvidia-docker run --rm -it --entrypoint /bin/bash terra-jupyter-gpu-dev-base:0.0.1` and then, inside the container, I run `conda install pytorch torchvision torchaudio cudatoolkit=11.3 -c pytorch` to install pytorch, then I can do `python -c "import torch; print(torch.cuda.is_available())"` and I get `True`
    - I can also successfully run `nvidia-smi`
    - This means the GPU is ready to roll
- Also made a Terra Cloud Environment successfully (with a GPU), and all the Jupyter stuff still works as above
    - `nvidia-smi` also works in Terra, meaning the GPU is good to go

image

I always build GPU-enabled images from the Nvidia docker image catalog (i.e. `FROM nvcr.io/nvidia/cuda:11.3.1-base-ubuntu20.04`).  They're great, and they're way smaller than the google deep learning images.  And this way, neither pytorch nor tensorflow is pre-installed.  People can build an image on top of this that installs either of those things, but they won't have stuff they don't need.

---------------

NOTE: the two new images in this PR are identical except for the `FROM` first line.  This could easily be refactored so that there's only one Dockerfile and two separate `build_docker.sh` files.  (It only works because the Nvidia base image is also ubuntu:20.04)

Qi77Qi:
```
@Qi77Qi is that something other than these?
https://github.com/sjfleming/terra-docker/blob/186a8b7225a04a2a704f26bbb939fdf4bd3e03b8/terra-jupyter-dev-base/Dockerfile#L140-L143
```

Is that actually working? you don't have `scripts` dir in `terra-jupyter-dev-base` to begin with...


Regardless tho, I think we're not planning to advertise these images for other users any time soon (if at all)...so it's really just for your own use case for now..If things are working for you, I don't mind merging them as is (adding a comment like "this image is not officially supported, use at your own risk" might be good in case others start using them). Bear in mind, it's additional work to add automation for building these 2 images, so you'll need to build them manually until we figure out if this is something we want to support

sjfleming:
Hi @Qi77Qi , yes, the reason I don't need to have the `scripts` copied into the same `terra-jupyter-dev-base` directory is that in the build script `build_docker.sh`, I am setting the "build directory" to the `terra-jupyter-base` directory, which contains the scripts, so docker can `COPY` them during the build:
https://github.com/sjfleming/terra-docker/blob/c5891655b6821b00e70e77e5b28f01c287bdfd71/terra-jupyter-dev-base/build_docker.sh#L3
it's the `../terra-jupyter-base` part at the end that sets the build directory.

That sounds good to me!  I'm sure building manually is fine for now.  If anybody else ends up using these images in the future, then maybe automation could be revisited.

Qi77Qi:
ah I see..that makes sense...whenever you're ready, feel free to put this to non draft

sjfleming:
Okay thanks Qí!  I will do a little bit of cleanup:
- refactor the GPU image to use the non-GPU Dockerfile with an input argument for the base image (since they are identical currently)
- use two requirements files for the python package installs, and pin all the versions of packages
  - `requirements.txt` that has all the "normal" python packages
  - `requirements_gcc.txt` that has the python packages that require gcc in order to build them (`firecloud` and `terra-notebook-utils`)
    - also explore whether we can get around these gcc requirements in the first place by doing a conda install instead of pip install?

sjfleming:
Turns out I was not including the google cloud SDK, but I realize that's 100% necessary.  (I'm guessing it's part of the tensor flow base image, so I didn't realize I was missing it at first, since it's not explicit in the "base" Dockerfile.)

That will increase the size of the image a bit.

sjfleming:
I am a bit confused about how cloud file syncing (syncing the notebook ipynb file to the project's google bucket) was working if my image did not have `gsutil`.  I guess there's another image running on the VM that takes care of that kind of copy operation?

sjfleming:
With `gsutil` the non-gpu image is now 1.96 GB

Feel free to experiment with it here (I've renamed it)
```
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-base:0.0.1
```

sjfleming:
And the GPU-enabled image is 2.02 GB

It is hosted here temporarily
```
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-gpu-base:0.0.1
```

(Of course this doesn't have `pytorch` or `tensorflow` installed, so that's why it seems small.)

sjfleming:
I installed `gsutil` to the home directory... that turned out to be a bad idea, since I think everything in there gets clobbered when the VM gets set up.  Let me fix that.

sjfleming:
It seems like `$PATH` is getting overwritten by some step in this process.  When I run an image locally, I see the `$PATH` I have specified in my docker build:
```
/opt/gcloud/google-cloud-sdk/bin:/opt/conda/bin:/home/jupyter/.local/bin:/home/jupyter/packages/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
But then when I actually use the image to start a Cloud Environment in Terra, I see something different.  It is always see the exact same `$PATH`, regardless of my Dockerfile, and it is the same for official images as well, like the GATK image:
```
/opt/conda/bin:/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/jupyter/.local/bin:/home/jupyter/packages/bin
```
@Qi77Qi is this expected behavior?

(I'm trying to fix a `gsutil command not found` error I am seeing in a Cloud Environment, and it seems to come down to the above problem.  For now, I will try to sidestep this by installing somewhere on that fixed PATH, if I can figure out how)

sjfleming:
Tested the images, and they seem to work, but I might have found one issue:
when I try 
```python
import os

os.environ['WORKSPACE_BUCKET']
```
I see `'.'`
while for other Cloud Environments I have built previously with an official Terra image, I see something more reasonable that's an actual bucket path.

**Where / how does the WORKSPACE_BUCKET environment variable get set?**

Other environment variables all seem to be okay, for instance
```python
os.environ['WORKSPACE_NAMESPACE']
```
`'bayer-pcl-cell-imaging'`
is correct

-------------------

This is now ready for review @Qi77Qi @jdcanas 

There are two new images included in this PR:
```
terra-jupyter-minimal-base
terra-jupyter-minimal-gpu-base
```
each can be built using the `build_docker.sh` script in the respective folder.

sjfleming:
After these last couple optimizations, `terra-jupyter-minimal-base` is 1.23 GB and `terra-jupyter-minimal-gpu-base` is 1.29 GB.

sjfleming:
Hey @thouis , it certainly didn’t die from my perspective.  But it looks like we haven’t made any progress toward merging this PR or making the minimal base images “official”.  I’d still love to see it happen.  Maybe your interest will help revitalize the effort.

sjfleming:
If you want, feel free to try

`us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-gpu-base:0.0.1`

from this PR.  That’s what I’m using now as the base for images I build for Terra.  (As with any custom image, I find it’s necessary to extend the default “timeout” for Cloud Environment creation to much longer than 10 mins.)

The only lingering problem is this (https://github.com/DataBiosphere/terra-docker/pull/346#issuecomment-1234767388) but I’ve been able to work around it.  I think I would need help from the experts to solve the issue.

thouis:
I spent some time yesterday building an updated version of the gpu base
image (cuda 11.8 base image, python 3.8 - I needed 3.8 for tensorqtl).  I
was able to use it for starting up an environment without increasing the
timeout (I've never figured out why this does or doesn't happen, either).

What is still missing for making these more official?  Is #346 indicative
of deeper concerns?

Ray

sjfleming:
Hi Ray, I would also like to update this to cuda 11.8 and python 3.8, since python 3.8 is now needed for pytorch 2.0+, which is the ecosystem I work in.  It's been quite a while since I touched this.

If you have made those changes, I'd be happy to merge them into my branch if you wanted them to become part of this PR.  Or I could incorporate the changes on my side if that's easier.  Was it just a matter of using a different base image and then using a different miniconda version?  And everything else still worked?

So the [comment above](https://github.com/DataBiosphere/terra-docker/pull/346#issuecomment-1234767388) is, I believe, an isolated issue that doesn't indicate deeper concerns.  But that's me guessing!  I've tested out other stuff pretty extensively, just by using it on Terra, and everything seems to work just fine.  I think what's going on is that somewhere deep in the Leonardo or Welder codebase (which I'm just not familiar with at all), the environment variables in any image run on Terra are being overwritten, so that users can access things like the workspace name and the bucket.  I have no idea why it's not working as expected here... and just the bucket environment variable.  The workspace name environment variable still works!  It really seems to just be the bucket.  It does bug me though.  But I can't figure it out myself.

Somebody from the Interactive Analysis DSP team please correct me if I'm wrong!  @rtitle do you know how these environment variables get set?

thouis:
My changes were:

Use these instead:
`nvcr.io/nvidia/cuda:11.8.0-base-ubuntu20.04`
`Miniconda3-py38_23.3.1-0-Linux-x86_64.sh`

For cromshell, I had to change to this (tags have changed?):
`git clone --branch 2.0.0 --single-branch https://github.com/broadinstitute/cromshell.git $HOME/cromshell`

Also, just after installing Miniconda, I install mamba (much faster drop-in replacement for conda)
`conda install -y mamba -c conda-forge`

And added this at the end (because mamba was complaining about compatible versions)
`RUN pip3 install -U requests`

Then I was able to use this to get torch:
`RUN mamba install pytorch pytorch-cuda=11.8 -c pytorch -c nvidia`

thouis:
The reason to install right away is that it can be difficult or impossible
to install after other things get added to the conda environment.  But I
understand the concern.

One could potentially use micromamba instead of conda (
https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) if the
goal is minimalism, but I think that's a more fraught choice if the goal is
to make this a base image for other environments.

sjfleming:
Alright @thouis , I have taken your advice, thank you for it.  And actually I went ahead and upgraded to python 3.10 since that's what all the other Terra docker images are using now (I think).  I did also install `mamba`.  Package versions across the board are now pinned to more up-to-date versions as well.

The images are currently hosted here, if anyone is interested:
```
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-base:0.0.2
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-gpu-base:0.0.2
```
sizes have increased slightly:
```
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-gpu-base   0.0.2       1.94GB
us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-base       0.0.2       1.8 GB
```

These images can successfully create a Terra Cloud Environment that seems to function normally (still barring the comment above about missing the environment variable WORKSPACE_BUCKET).

sjfleming:
Compressed sizes:
```
terra-jupyter-minimal-base         674.4 MB
terra-jupyter-minimal-gpu-base     724.8 MB
```

nick-youngblut:
Any status updates on merging `terra-jupyter-minimal-base`? My group also needs Python 3.8 (as with @thouis ) for some specific bfx software. 

If anything, we could probably modify `us.gcr.io/broad-dsde-methods/terra-jupyter-minimal-base:0.0.2` from [this PR](https://github.com/DataBiosphere/terra-docker/blob/e27c4ca6a94c0e76390acd5cedbcb8f2f1aa7ce7/terra-jupyter-minimal-base/Dockerfile) in order to use Python 3.8 instead of Python 3.10, but it would be great if there was more support for each major Python version.

Also, it was a bit of a treasure hunt to find this PR, since this info isn't in the [Terra-Bio docs](https://support.terra.bio/hc/en-us/sections/360003924192-Working-with-Containers-Docker-), AFAIK.

sjfleming:
@nick-youngblut I think the official effort toward making this happen has moved to this PR:
https://github.com/DataBiosphere/terra-docker/pull/463

PR #347, opened 2022-08-12
[SUP-496] Reverted AOU changelog and version.

n8landolt:


Qi77Qi (commented):
https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json#L146 needs to be reverted if we intend to revert

Qi77Qi (approved):

yonghaoy:
@Qi77Qi @n8landolt 
We had few conflicts issues before when reusing the a reversed version number. 
I think we should flag any version created in this PR as "do not use" in respective changelogs
Example: https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-aou/CHANGELOG.md#214---2022-06-03t183522582705z

Qi77Qi:
@yonghaoy  2.1.9 version was never published...which is why in this PR, we're reverting that in the changelog

yonghaoy:
> @yonghaoy 2.1.9 version was never published...which is why in this PR, we're reverting that in the changelog

ah, thanks for clarifying!

PR #348, opened 2022-08-17
[IA-3482] Adding fixes to the generate_packages_docs script

xuscrbw:
A few of these commands were failing. Adding a library() call to import the library, which then allows this to run. Also disabling a pip version check to avoid a warning printing in the list of python images. This is so the mapping works well


Qi77Qi (approved):
👍 🎉

PR #366, opened 2022-10-19
Update terra-jupyter-hail to Hail 0.2.104

danking:


Qi77Qi:
@xuscrbw while you're working on the base image upgrade. Can you also bump hail to latest as well. Thanks!

PR #319, opened 2022-05-02
Add updated dockerfile for Bioc3.15, R-4.2.0

nturaga:
@gpcarr This should the PR which updates `terra-jupyter-r` image.

Note: This PR includes a Bioconductor version update, so a minor version bump on version number is required for the image for the ammonite script to work. i.e the CHANGELOG should indicate

```
diff --git a/terra-jupyter-r/CHANGELOG.md b/terra-jupyter-r/CHANGELOG.md
index 80a545b..f15925d 100644
--- a/terra-jupyter-r/CHANGELOG.md
+++ b/terra-jupyter-r/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 2.1.0 - 2022-05-02
+
+- Update Bioconductor to 3.15.0
+  - Install R 4.2.0
+  - Update all inheriting image 
+
+Image URL: `us.gcr.io/broad-dsp-gcr-public/terra-jupyter-r:2.1.0`

PR #373, opened 2022-11-22
Change Bioconductor version for Jupyter

almahmoud:
My understanding from document is that this is handled by Broad, but I started the PR for tracking.

Friends with https://github.com/DataBiosphere/terra-docker/pull/372


Qi77Qi:
see https://github.com/DataBiosphere/terra-docker/pull/319.. there're a few misc files need to be updated..

almahmoud:
I used `amm` to update the versions. Can you please specify what "misc files" you are referring to?

Qi77Qi:
oh I thikn the misc files are actually updated by the amm script already..so I think it's all good in that regard..the remaining thing is try to figure out why the tests are failing

PR #375, opened 2022-11-30
[SU-225] Add base JupyterLab Docker image (draft)

MatthewBemis:


terra-jupyterlab-python/Dockerfile
```diff
@@ -0,0 +1,76 @@
+FROM us.gcr.io/broad-dsp-gcr-public/terra-jupyter-base:1.0.11
```
Qi77Qi, line 1: we can maybe have a meeting with Alessandro tgt to talk about requirements..but I'm not sure if we want to have a separate jupyterlab base image cuz we don't wanna maintain a suite of separate jupyterlab-r, jupyterlab-aou etc

MatthewBemis: That's reasonable. One reason that I wanted to use a separate image is that it gives us the space to update dependencies and be more experimental with JupyterLab without having to worry about breaking things for existing Jupyter users. But that does come at the cost of extra maintenance, even if it's just a short-term approach for the preview phase.

My Leo approach also relies on an env var in this new image to determine that the user is spinning up JupyterLab as a tool instead of Jupyter, but I suppose Terra UI could add a runtime label during runtime creation that gives Leo the same information.

Do you have a preference? I definitely don't want to create a maintenance burden for IA :)

Qi77Qi: >that it gives us the space to update dependencies and be more experimental with JupyterLab without having to worry about breaking things for existing Jupyter user
for sure, this is where I think a meeting with Alessandro to clarify some product requirements will be helpful.

> determine that the user is spinning up JupyterLab as a tool instead of Jupyter
if we're merging this with the existing stack, I think the image should be able to support both jupyter and jupyterLab...And then UI can specify whether user wants to spin it up as jupyter notebook or jupyterlab somehow

PR #382, opened 2022-12-14
Update terra-jupyter-hail to Hail 0.2.107

danking:


dmohs (approved):
Would prefer formatting changes in a separate PR, but appreciate seeing them.

danking:
Ah, sorry, I think this is the default format of jq. I can dig a bit right now and see if I can change the default to conform to this multiple line representation of empty objects.

danking:
`jq` does not appear to support any configuration of formatting. Are y'all using a JSON formatting tool? I'm happy to add it to the CI/CD pipeline which generates these PRs.

PR #386, opened 2023-01-30
Update dockerfile

MilanParikh:
Adding "&&" to two lines to prevent the debian update commands being added to the sources list. This is causing errors when installing/updating packages.


n8landolt (approved):

jdcanas (approved):
We're working on debugging the github action failure...

PR #387, opened 2023-01-31
AoU RStudio

mklarqvist:
RStudio Docker with ubuntu tasty with nginx + wondershaper bandwidth throttling


yonghaoy (commented):
Thanks! This is awesome. We might need some Leo folks help to review this PR. 
@jdcanas or @rtitle might be able to help while Qi is out. 
Few things that on top of my head: 

1. We would need to update this [build script](https://github.com/DataBiosphere/terra-docker/blob/master/build.sh) which does build (then push later [in separate job](https://github.com/DataBiosphere/terra-docker/blob/master/jenkins/Jenkinsfile)) for us
2. Shall we add some smoke test like our [current Jupyter image](https://github.com/DataBiosphere/terra-docker/blob/master/.github/workflows/test-terra-jupyter-aou.yml)? 
3. I know [config.json](https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json) is needed to track all images, but I am not sure how that can be used and what shall we do here.
4. (for later step): We need to create an AoU  RStudio APP and point it to our new images once we have that: https://github.com/DataBiosphere/terra-app/blob/main/apps/rstudio/app.yaml and need to make sure command and args in that file compatible with the Dockerfile we have (we had an issue before when that file uses 8002 while Dockerfile uses 8787, Qi and Rob helped us resolved that issue).

mklarqvist:
Great!  Feel free to take over or build on this PR

yonghaoy:
Also added InfoSec to review

jdcanas:
> Thanks! This is awesome. We might need some Leo folks help to review this PR. @jdcanas or @rtitle might be able to help while Qi is out. Few things that on top of my head:
> 
> 1. We would need to update this [build script](https://github.com/DataBiosphere/terra-docker/blob/master/build.sh) which does build (then push later [in separate job](https://github.com/DataBiosphere/terra-docker/blob/master/jenkins/Jenkinsfile)) for us
> 2. Shall we add some smoke test like our [current Jupyter image](https://github.com/DataBiosphere/terra-docker/blob/master/.github/workflows/test-terra-jupyter-aou.yml)?
> 3. I know [config.json](https://github.com/DataBiosphere/terra-docker/blob/master/config/conf.json) is needed to track all images, but I am not sure how that can be used and what shall we do here.
> 4. (for later step): We need to create an AoU  RStudio APP and point it to our new images once we have that: https://github.com/DataBiosphere/terra-app/blob/main/apps/rstudio/app.yaml and need to make sure command and args in that file compatible with the Dockerfile we have (we had an issue before when that file uses 8002 while Dockerfile uses 8787, Qi and Rob helped us resolved that issue).

So, if you want to hook into our existing build system, you'll need to put the needed syntax to build the docker image in `build.sh` as @yonghaoy says. I see you have a docker compose, but I'm a little confused because there are 2 Dockerfiles in the sub folder?

1: Regardless, `build.sh` should be able to handle any alternate syntax needed for building, and the existing job should be able to build it.
2: Smoke tests would be ideal!
3: Config.json is how the job mentioned in 1) finds the image

dinvlad:
From AppSec perspective, we're not yet switching to Wondershaper running in a separate container right? And is it correct that there's a plan for doing that in the future?

aou/aou-rstudio-bioconductor/rstudio/scripts/init_userconf.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/with-contenv bash
+# shellcheck shell=bash
+
+## Set defaults for environmental variables in case they are undefined
+DEFAULT_USER=${DEFAULT_USER:-rstudio}
+USER=${USER:=${DEFAULT_USER}}
+USERID=${USERID:=1000}
+GROUPID=${GROUPID:=1000}
```
dinvlad, line 8: These are not user-overridable right?

aou/aou-rstudio-bioconductor/rstudio/Dockerfile
```diff
@@ -0,0 +1,137 @@
...
+
+ENV DISABLE_AUTH=true
+ENV ROOT=false
+
+# Restrict download speed to 8Mbit/s = 1MB/s and upload speeds to 4Mbit/s = 512KB/s
+# Wondershaper will be executed as a different user compared to the rstudio user
+# in /init.  Should not be possible to reset Wondershaper to unlimited speeds or delete
+# the Wondershaper executable.
+# TODO: Add VPC ip addresses to whitelist
+CMD /bin/bash -c "wondershaper eth0 8096 4048 && nginx -g 'daemon on;' && /init"
```
dinvlad, line 137: Are we using the default Nginx config file?

dinvlad, line 137: Also, we might want to use ENTRYPOINT with a script that uses [su-exec](https://github.com/ncopa/su-exec) or [gosu](https://github.com/tianon/gosu) that launches wondershaper, Nginx and /init as different users, such that there's no shell script running at PID 1.

yonghaoy:
Question from @jdcanas, Why are there 2 docker images, without a top-level one though? Is the intent that the docker compose builds the rstudio folder?

My understanding to make this work with Custom App, there should be docker images that has everything. Leo Custom App does not support multi-containers

yonghaoy:
How does nginx  work with Leo GKE.
Leonardo will install nginx ingress controller for Custom App: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/resources/reference.conf#L485L495
We need to make sure it works with Leo helm chart...

mklarqvist:
> Question from @jdcanas, Why are there 2 docker images, without a top-level one though? Is the intent that the docker compose builds the rstudio folder?
> 
> My understanding to make this work with Custom App, there should be docker images that has everything. Leo Custom App does not support multi-containers

Answered above https://github.com/DataBiosphere/terra-docker/pull/387#issuecomment-1412672373

aou/aou-rstudio-bioconductor/rstudio/scripts/default_user.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -e
+
+DEFAULT_USER=${1:-${DEFAULT_USER:-"rstudio"}}
```
yonghaoy, line 4: what is default user?  
The RStudio we use will disable auth: 
https://github.com/DataBiosphere/terra-app/blob/main/apps/rstudio/app.yaml#L10L16
The file above is the RStudio config Leo will install for us when we call Leo.

mklarqvist: If you want to dig into the details of how the rocker dockers are setup it's better you ask them
https://github.com/rocker-org/rocker-versioned2/blob/23b6961dcc187b7290b35ae4434180dfb1fa7f24/scripts/default_user.sh

All the script files are copied here because we had to change the base image.

mklarqvist: And we do exactly the same
```
ENV DISABLE_AUTH=true
ENV ROOT=false
```

mklarqvist: The contents of this PR is identical to `us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor:3.15.3` with the exception that we have a different base image with wondershaper (hence why we have to copy-paste their scripts) and we add 25 more packages in the end.
That's it.

aou/aou-rstudio-bioconductor/rstudio/scripts/init_userconf.sh
mklarqvist: Correct, RStudio remaps/regroups UID and GID at runtime inside `/init`

aou/aou-rstudio-bioconductor/rstudio/Dockerfile
mklarqvist: > From AppSec perspective, we're not yet switching to Wondershaper running in a separate container right? And is it correct that there's a plan for doing that in the future?

Correct.  Hence this annoying PR where we copy-paste all those other images together with a different base.

mklarqvist: Services started before `/init` will have a different UID and GID. `/init` (which is a script) takes creates a new UID and GID and sets permissions for the RStudio user.  The user should not be able to stop, remove, or modify those running services.
In my tests, this appears to be true.

mklarqvist:
Turns out we can rewrite the entire thing with multistage builds to lift over what we need from another image instead of redoing all the work.

mklarqvist:
Found workarounds to to extend `us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor` in a multi-stage build resulting in ~1000 less lines of code.

build.sh
```diff
@@ -42,7 +42,7 @@ docker push $GCR_IMAGE_REPO/$IMAGE_DIR:$TAG_NAME
 docker push $GCR_IMAGE_REPO/$IMAGE_DIR:$VERSION
 docker push $GCR_IMAGE_REPO/$IMAGE_DIR:latest
 
-if [[ $IMAGE_DIR = "terra-jupyter-aou" ]]; then
```
yonghaoy, line 45: nit: we don't need this change because we also migrated from dockerhub to gcr.

config/conf.json
```diff
@@ -144,6 +144,22 @@
                 "include_in_custom_gce" : true
             }
         },
+        {
+            "name" : "terra-aou-bioconductor-tidyverse-3.16.0",
```
yonghaoy, line 148: Those names is a bit confusing. I saw 
`terra-aou-bioconductor-tidyverse-3.16.0` 
`aou-rstudio-bioconductor`
and `aou-rstudio-bioconductor-3.16.0`
I think the naming conversion is dir name is the same as image name. Suggest to make them be consistent with other images. 
Also, suggest to remove 3.16.0 from image name as it is indicated in version.

config/conf.json
```diff
@@ -144,6 +144,22 @@
...
+                "r"
+            ],
+            "packages" : {},
+            "version" : "3.16.0",
+            "automated_flags" : {
+                "include_in_ui" : false,
+                "generate_docs" : false,
+                "build" : true,
+                "include_in_custom_dataproc" : true,
+                "include_in_custom_gce" : true
```
yonghaoy, line 160: Those two should be false. @jdcanas  to confirm

yonghaoy:
> 

Not sure how to address my above comment: dir & image name. If we agree on those names,  I can take it over and polish it a bit.

config/conf.json
mklarqvist: 👍  I have no idea what they are used for.  I literally copied the section from the other AoU image 
https://github.com/DataBiosphere/terra-docker/blob/a313154bf0f26a1df5c3c5e61bb9c536016a75ed/config/conf.json#L130:L146

aou/aou-rstudio-bioconductor/rstudio/Dockerfile
dinvlad: Ok - I'd still suggest to use su-exec or gosu for all 3 processes, so that they don't run under /bin/bash

aou/aou-rstudio-bioconductor/rstudio/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Run wondershaper to limit network bandwidth to 8mbps down and 4mbps up as
+# a different user than RStudio user which is set in the /init script.
+su-exec root wondershaper eth0 8096 4048 && nginx -g 'daemon on;'
+
+# Init will handle it's own UID and GID mapping
+/init
```
dinvlad, line 8:
```suggestion
su-exec $UID:$GID /init # please replace with appropriate UID/GID
```

config/conf.json
jdcanas: Yeah, these should can false. This flag causes them to be cached in the VM image created for VMs used by Terra UI.

config/conf.json
jdcanas: Yeah, the image dir should not have a version, nor should this name. Versioning is handled elsewhere. Also recommend not including any versioning in the dockerfile for the same reason

aou/aou-rstudio-bioconductor/rstudio/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Run wondershaper to limit network bandwidth to 8mbps down and 4mbps up as
+# a different user than RStudio user which is set in the /init script.
+su-exec root wondershaper eth0 8096 4048 && nginx -g 'daemon on;'
```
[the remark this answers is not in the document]
mklarqvist: 👍  Added this

aou/aou-rstudio-bioconductor/rstudio/entrypoint.sh
mklarqvist: This still needs to run as `root` since it runs `S6` in `/init` and sets `UID` and `GID` in there as a non-root user called `rstudio`

mklarqvist: Added some comments to explain

config/conf.json
mklarqvist: Set them to `false`

config/conf.json
mklarqvist: Updated names

aou/aou-rstudio-bioconductor/rstudio/scripts/default_user.sh
mklarqvist: default user is `rstudio` and is run as a non-root user with defaults `1000:1000` in `S6`

aou/aou-rstudio-bioconductor/rstudio/Dockerfile
mklarqvist: We don't actually need nginx to be running.  We just needed the kernel modules and settings and configs from that docker image

mklarqvist:
We don't need to run the nginx service.  We were just after the kernel modules and settings etc from the `ubuntu/nginx` image.
```bash
rstudio@aou:~$ speedtest-cli
Retrieving speedtest.net configuration...
Testing from Google Cloud (34.135.35.254)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Sentco.net (Falls City, NE) [134.90 km]: 26.497 ms
Testing download speed................................................................................
Download: 7.19 Mbit/s
Testing upload speed......................................................................................................
Upload: 4.56 Mbit/s
```

mklarqvist:
Using `tc` directly with
```
# Restrict uploads to 4096 kbit/s (0.512 MB/s) and unlimited download
tc qdisc add dev eth0 root tbf rate 4096kbit latency 50ms burst 1540000
```
we can simplify the build even more.  Setting default download speeds to unlimited and upload to 512 kb/s.

```
rstudio@aou:~$ speedtest
Retrieving speedtest.net configuration...
Testing from Google Cloud (34.135.35.254)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Skywave (West Point, NE) [96.34 km]: 29.623 ms
Testing download speed................................................................................
Download: 2088.90 Mbit/s
Testing upload speed......................................................................................................
Upload: 5.64 Mbit/s
```

terra-aou-rstudio/rstudio/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Restrict uploads to 4096 kbit/s (0.512 MB/s) and unlimited download
+tc qdisc add dev eth0 root tbf rate 4096kbit latency 50ms burst 1540000
+
+# Init must be run as root and will internally handle mapping the end-user
+# to UID and GID mapping using S6-overlay
+# https://github.com/just-containers/s6-overlay
+# using S6
+# https://skarnet.org/software/s6/overview.html
+/init
```
dinvlad, line 13:
```suggestion
su-exec root /init
```
Sorry, this is what I meant - every process called from the entrypoint script should be prepended with `su-exec root`, so that the entrypoint script exits after forking these processes.

terra-aou-rstudio/rstudio/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e # catches errors
+
+# Restrict uploads to 4096 kbit/s (0.512 MB/s) and unlimited download
+tc qdisc add dev eth0 root tbf rate 4096kbit latency 50ms burst 1540000
```
dinvlad, line 6: also please double-check that end-user is unable to call `tc`

jdcanas (approved):
Haven't tested this, if it works with `build.sh` I'd say its good to go

terra-aou-rstudio/rstudio/entrypoint.sh
mklarqvist: Removed permissions to even view `tc` and others in `sbin`

```
rstudio@aou:~$ tc
bash: tc: command not found
rstudio@aou:~$ speedtest
Retrieving speedtest.net configuration...
Testing from Google Cloud (34.135.35.254)...
Retrieving speedtest.net server list...
Selecting best server based on ping...
Hosted by Binary.Net (Omaha, NE) [12.44 km]: 29.904 ms
Testing download speed................................................................................
Download: 2045.38 Mbit/s
Testing upload speed......................................................................................................
Upload: 5.71 Mbit/s
```

terra-aou-rstudio/rstudio/entrypoint.sh
mklarqvist:
```
Tasks:   7 total,   1 running,   6 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.2 us,  0.0 sy,  0.0 ni, 99.7 id,  0.2 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem :   7946.7 total,   1589.5 free,    481.1 used,   5876.1 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used.   7159.1 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                          
    313 rstudio   20   0  923360 149224  79628 S   0.3   1.8   0:00.86 rsession                                         
      1 root      20   0     204      4      0 S   0.0   0.0   0:00.02 s6-svscan                                        
     37 root      20   0     204      4      0 S   0.0   0.0   0:00.00 s6-supervise                                     
    267 root      20   0     204      4      0 S   0.0   0.0   0:00.00 s6-supervise                                     
    270 rstudio+  20   0  258308  17800  15372 S   0.0   0.2   0:00.15 rserver                                          
    339 rstudio   20   0    8668   5420   3736 S   0.0   0.1   0:00.01 bash                                             
    343 rstudio   20   0   10376   3916   3344 R   0.0   0.0   0:00.00 top                                              
```

terra-aou-rstudio/rstudio/entrypoint.sh
dinvlad: cannot even do `/sbin/tc` right?

dinvlad (approved):

mklarqvist:
I can successfully run `build.sh`
```
...
+ docker image rm -f us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:3.16.0
Untagged: us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:3.16.0
+ docker image rm -f us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:c1662b7
Untagged: us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:c1662b7
+ echo 'Successfully completed build script for terra-aou-rstudio'
Successfully completed build script for terra-aou-rstudio
```

yonghaoy:
> I can successfully run `build.sh`
> 
> ```
> ...
> + docker image rm -f us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:3.16.0
> Untagged: us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:3.16.0
> + docker image rm -f us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:c1662b7
> Untagged: us.gcr.io/broad-dsp-gcr-public/terra-aou-rstudio:c1662b7
> + echo 'Successfully completed build script for terra-aou-rstudio'
> Successfully completed build script for terra-aou-rstudio
> ```

I was also trying to fix this error. The reason is like what I mentioned above, we need to make image name be consistent with dir name. 
I haven't heard your naming preference, so I had a local version rename that entier dir to terra-aou-rstudio-bioconductor.
But what you have also LG, I will revert mine. 
And I was trying to take over your PR and made few other modification, but I do not have push permission on your folked repo..

mklarqvist:
Added the changes from @yonghaoy .  I have no idea how you merge PRs so I copy pasted

PR #388, opened 2023-02-03
Yyu rstudio

yonghaoy:


yonghaoy:
Changes that I made based on @mklarqvist PR: 

1. Add readme
2. Add Jenkins job to allow it to build via Leo
3. Remove the cloudbuild.yml because it can be covered by existing workflow
4.

build.sh
```diff
@@ -42,7 +42,7 @@ docker push $GCR_IMAGE_REPO/$IMAGE_DIR:$TAG_NAME
 docker push $GCR_IMAGE_REPO/$IMAGE_DIR:$VERSION
 docker push $GCR_IMAGE_REPO/$IMAGE_DIR:latest
 
-if [[ $IMAGE_DIR = "terra-jupyter-aou" ]]; then
+if [[ $IMAGE_DIR = "terra-jupyter-aou" ]] ; then
```
jdcanas, line 45: You'll probably want to add the added image dir here so we handle AOU images the same way

jdcanas:
(Reminder to myself) Let's make sure we run the smoke test workflow once this isn't a draft anymore!

PR #391, opened 2023-02-09
[IA-4036] Updated base dockerfile and appropriate documents.

n8landolt:
For reviewers:
There are a lot of files changed, but they can be easily categorized.
The .yml file changes are all updating the authentication key for Github actions.
The Dockerfiles (Other than terra-jupyter-base) have spacing changes.
And smoke_test.py had a test removed for the time being. Follow up ticket: https://broadworkbench.atlassian.net/browse/IA-4061

### DIFF
--- a/terra-jupyter-base/Dockerfile
+++ b/terra-jupyter-base/Dockerfile
@@ -26,8 +26,9 @@ RUN rm /etc/apt/sources.list.d/cuda.list \
     && curl -L -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-keyring_1.0-1_all.deb \
     && sudo dpkg -i cuda-keyring_1.0-1_all.deb
 
-RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
-    apt-get update && apt-get install -yq --no-install-recommends \
+RUN wget -O - -q https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+
+RUN apt-get update && apt-get install -yq --no-install-recommends \
     # gnupg requirement
     dirmngr \
     gnupg \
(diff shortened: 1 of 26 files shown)


terra-jupyter-base/Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN rm /etc/apt/sources.list.d/cuda.list \
     && curl -L -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-keyring_1.0-1_all.deb \
     && sudo dpkg -i cuda-keyring_1.0-1_all.deb
 
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+RUN sudo curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
```
danking, line 30: the sudo on curl should be unnecessary (curl itself doesn't need root privileges, just apt-key)

n8landolt: I THINK we're already root as well, so I'm not sure we even need any `sudo`s 😆

terra-jupyter-python/tests/smoke_test.py
```diff
@@ -25,31 +29,30 @@ def test_pandas():
     }
   )
   
-def test_pandas_gbq():
-  import pandas_gbq
+# def test_pandas_gbq():
```
LizBaldo, line 32: Instead of commenting out, could you you mark this test as skipped with a decorator like this?
`@pytest.mark.skip(reason="issues passing in credentials, see IA-4061")`

LizBaldo (approved):
LGTM, looks like the dockerfiles and jsons conf got reformated, but as long as it is still working, should not be an issue?

If you do not mind skipping the test with a reason using pytest decorator instead of commenting out before merging that would be great!

yonghaoy:
@n8landolt do we have a ticket to document cromshell upgrade? AoU is using cromshell-alpha, and the upgrade will break us

PR #399, opened 2023-02-28
[RW-9461] Install and use wondershaper when starting app. Also fix Leo image publishing script

yonghaoy:
Images is pushed to GCR successfully: https://console.cloud.google.com/gcr/images/broad-dsp-gcr-public/us/terra-rstudio-aou?project=broad-dsp-gcr-public

- Update and set wondershaper. I verified this by using this image. 
- Fix build script and add README.
- Fix GHA branch. But I am not able to test it now. I am not sure if it requires a Github SA like this: https://cs.github.com/DataBiosphere/terra-cloud-resource-lib/blob/b8931f7951d3ee0df23e2b056630c491a9fad8e3/.github/workflows/tag-publish.yml?q=%22checkout%40v2%22#L63
I might need to come back and update token part 

However, the GHA failed because it is not compatible with the current generate python package script. 

Also update README file about how to use this GHA


yonghaoy:
Hey @jdcanas I updated this PR to include wondershaper changes. PTAL
Here is the summary: 

- Update and set wondershaper. I verified this by using this image. 
- Fix build script and add README.
- Fix GHA branch. But I am not able to test it now. I am not sure if it requires a Github SA like this: https://cs.github.com/DataBiosphere/terra-cloud-resource-lib/blob/b8931f7951d3ee0df23e2b056630c491a9fad8e3/.github/workflows/tag-publish.yml?q=%22checkout%40v2%22#L63
I might need to come back and update token part

PR #405, opened 2023-03-14
Update terra-jupyter-aou to Hail 0.2.111

danking:


danking:
I'm happy to make the code change, but I'm not comfortable making the decision as I'm not responsible for these images. Presumably folks expect BOLT-LMM to be present in their images?

danking:
Hail 0.2.112 is coming soon.

danking:
Oh, heh, Alkes' group. They had someone recently rack up a big charge by egressing from their bucket. They'll soon be part of the S3 open data program and y'all can choose to pull from there once that happens. I'm actually the one responsible for copying their data to S3 and I don't have time to do that until next week.

PR #394, opened 2023-02-17
[SU-293] Fix Jupyter Lab build message on start

MatthewBemis:

PR #410, opened 2023-03-16
[no ticket] Fix docker version because base image v13 does not exist in GCR

yonghaoy:
It might be a typo in https://github.com/DataBiosphere/terra-docker/pull/394, and it makes `python, gatk, r, aou` images fail to build and push.


jdcanas (approved):

MatthewBemis (approved):

yonghaoy:
@jdcanas shall I manually kick off push image to GCR GHA or Leo has another Jenkins job that will run automattly?

jdcanas:
if there's not an image built and pushed for these images, the CI will do it. If an image exists, you'll need to bump the versions for the images here and run `updateVersions.sc`

PR #414, opened 2023-04-18
[IA-4148] updated the LargeFileManager import in jupyter_delocalize.py

n8landolt:
https://broadworkbench.atlassian.net/browse/IA-4148

This PR updates the import from notebook to Jupyter Server. 

It seems the package naming changed with the new DSVM image.


LizBaldo (approved):
Good catch! :shipit:

PR #422, opened 2023-05-23
Update terra-jupyter-hail to Hail 0.2.117

danking:


LizBaldo:
@danking do you need 0.2.117 specifically or would 0.2.115 work? This does not work because we are on Python 3.7 (planning on upgrading to python 3.10 in about 3 weeks). So I am happy to release a new image with the most up to date version of hail that is compatible with 3.7 (aka hail 0.2.115), or we can wait a few weeks until the python upgrade is over. Please lmk!

danking:
These are just automatically generated PRs from our grease pipeline, there is no particular urgency from me. 😊

danking:
0.2.118 will land shortly and make new PRs. Closing.

PR #425, opened 2023-05-24
[RW-10154][RW-10159] Fix Cromshell alias and revert Hail upgrade

yonghaoy:
1: The newer Hail version requires higher version of Spark. Spark is intalled via Leonardo. Rollback Hail for now until Leo upgrade Dataproc version. 
2: Fix a typo of Cromshell alias.

### DIFF
--- a/config/conf.json
+++ b/config/conf.json
@@ -143,7 +143,7 @@
             "packages" : {
                 
             },
-            "version" : "2.1.20",
+            "version" : "2.1.21",
             "automated_flags" : {
                 "include_in_ui" : false,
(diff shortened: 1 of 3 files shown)


yonghaoy:
@Peter-Lavigne Not sure how to fix the R version issue in your Cromshell upgrade PR, so I make a fix just for AoU images to unblock our image upgrade

Peter-Lavigne (approved):

Peter-Lavigne:
Thanks! I'll keep investigating the R version issue.

PR #426, opened 2023-05-30
[IA-3730] Installing bcftools and bedtools in jupyter GATK image

LizBaldo:
JIRA ticket: https://broadworkbench.atlassian.net/browse/IA-3730

Note that the aou image already installs bcftools so I used the same installation commands, building the package from Github directly.Bedtools follows a similar approach


terra-jupyter-gatk/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## 2.2.11 - 2023-03-13T17:26:34.219509Z
```
mlilynolting, line 1: I think this version + timestamp needs an update

mlilynolting (approved):

PR #432, opened 2023-06-15
RW-9307 wondershaper docker image

tarekmamdouh:
I’ve created a wondershaper image and published it to this path https://pantheon.corp.google.com/gcr/images/broad-dsp-gcr-public/us/wondershaper?project=broad-dsp-gcr-public

I think we can still have this PR merged and keep the wondershaper image to use it in the future if needed, despite that we may be using GCE apps instead of GKE in the near future


tarekmamdouh:
Tested in GCE
image

yonghaoy (approved):

PR #444, opened 2023-07-26
IA-4362 remove java 8 for the gatk and aou images and update hail to 120

LizBaldo:
Forgot to also remove the installation of Java 8 for the AOU image, we are getting Java 11 form the new debian11 dataproc image in leonardo now, and it conflicts with the Java 8 that was previously installed on the image

Instead of bumping up the version, i am just going to override the existing one since we never released it

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -47,7 +47,7 @@ ENV PYTHONPATH $PYTHONPATH:/usr/lib/spark/python
 
 ENV PYSPARK_PYTHON=python3
 
-ENV HAIL_VERSION=0.2.119
+ENV HAIL_VERSION=0.2.120
 
 RUN find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
     && $JUPYTER_HOME/scripts/kernel/kernelspec.sh $JUPYTER_HOME/scripts/kernel /opt/conda/share/jupyter/kernels \
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -58,11 +58,11 @@ RUN find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
     && update-alternatives --install /etc/hadoop/conf hadoop-conf /etc/hadoop/conf.empty 100 \
     && update-alternatives --install /etc/hive/conf hive-conf /etc/hive/conf.dist 100 \
     && apt-get update \
-    && apt install -yq --no-install-recommends openjdk-8-jdk \
-        g++ \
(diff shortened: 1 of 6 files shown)


terra-jupyter-aou/Dockerfile
```diff
@@ -58,11 +58,11 @@ RUN find $JUPYTER_HOME/scripts -name '*.sh' -type f | xargs chmod +x \
     && update-alternatives --install /etc/hadoop/conf hadoop-conf /etc/hadoop/conf.empty 100 \
     && update-alternatives --install /etc/hive/conf hive-conf /etc/hive/conf.dist 100 \
     && apt-get update \
-    && apt install -yq --no-install-recommends openjdk-8-jdk \
-        g++ \
-        liblz4-dev \
-    # specify Java 8
-    && update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java \
+    # Note that starting on 2.1.x, dataproc VM also provides java 11, which is incompatible with 8. No need to install it in the container
+    # && apt install -yq --no-install-recommends openjdk-8-jdk \
```
Qi77Qi, line 62: does gatk work with java 11 now?

I vaguely remember gatk was the reason why we used java8

LizBaldo: Hail is built with java8, which was the original reason why we put this here, but it does work with Java 11 as well, as long as we don't mix and match java versions (I tested this with the terra hail image and it works fine)

lucymcnatt (approved):

LizBaldo:
I just heard from the hail team that they are fixing the bug that I reported for version 0.220, I will ask when they are planning on releasing it so we can use it 🤞

terra-jupyter-aou/Dockerfile
Qi77Qi: gatk also needs java...

we should make sure [this version](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-gatk/Dockerfile#L37) works with java 11

LizBaldo: Do you have a way to test the aou image on a BEE? So far we tested GATK on terra and it is working, but we don't have a good way to test the AOU image

LizBaldo:
@Qi77Qi I opened this PR thinking that AOU was using this image with a dataproc cluster, am I wrong?
If AOU is planning on using hail, how is it getting spark if not via the dataproc image?

If AOU is using the dataproc image then we also need to remove java 8 from theGATK image, and if we need to modify that, might as well upgrade GATK to 4.3.0

CC @lucymcnatt

LizBaldo:
FYI, I can confirm that the GATK image comes with Java 11 already so we should be good removing Java 8 👍

PR #445, opened 2023-07-26
Change legacy BiocJupyter to last 3.16 container

almahmoud:
`2.1.10` is the last version with R 4.2.2 and Bioc 3.16.
Since the latest container is Bioc 3.17 and R 4.3.0, it makes sense to leave the legacy at Bioc 3.16, rather than Bioc 3.17 with older python as is `2.1.11`.

### DIFF
--- a/config/legacy_static_images.json
+++ b/config/legacy_static_images.json
@@ -1,11 +1,11 @@
 [
   {
     "updated": "2023-06-23",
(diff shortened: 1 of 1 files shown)


LizBaldo (approved):

LizBaldo:
@almahmoud we have not yet released the new images with 3.10 in prod so it is the perfect time to sneak this in, are you good with me merging?

lucymcnatt (approved):

PR #450, opened 2023-08-15
Add some missing sysdeps for Bioc/R packages

almahmoud:
Some missing sysdeps surfaced in the new base container after it was changed for the python upgrade


almahmoud:
Since 3.18 is upon us and this PR never got merged, closing and adding commit to 3.18 PR

PR #463, opened 2023-10-11
[IA-4534] Leaner terra base docker image

LizBaldo:
JIRA Ticket: https://broadworkbench.atlassian.net/browse/IA-4534

More details can be found in the design doc here: https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/2842460165/2023-08-28+Terra+Docker+Refactoring+POC#Provide-a-leaner-set-of-terra-docker-images

The main changes are:
- Use a much smaller base cuda enabled ubuntu image (80mb vs. 8Gb for the gpu tensorflow one we are currently using)
- Use poetry to manage the python environment and it's dependencies

The goal is to make the base image small enough that it can be used by both custom image users AND provide deterministic python environments so developers don't end up in pip dependency hell every time we try to update the image

Note that I copied over the custom jupyter setup script, so for reviewing you should focus on the following files:
- README.md
- pyproject.toml
- Dockerfile

Given that this is a POC, building and publishing the image with GHA is outside of the scope of this PR and I will do it manually for now


terra-base/Dockerfile
```diff
@@ -0,0 +1,178 @@
...
+ENV JUPYTER_HOME /etc/jupyter
+ENV JUPYTER_KERNELSPEC_DIR /usr/local/share/jupyter
+# We need node >18 for jupyter to work
+ENV NODE_MAJOR 20
+
+#######################
+# Users Setup
+#######################
+# Create the jupyter user, add it to the `users` group and specify the home directory path
+ENV USER jupyter
```
cpate4, line 20: At this point, I believe your `USER` is `root`, and we are overwriting the env `USER` variable in the active environment with a different user name. This variable is actually overwritten further down this Dockerfile on this line `USER $USER` where we actively change the active user for the image.

LizBaldo: You are correct, this is my intention here :)

PR #472, opened 2023-11-07
Add BiocRStudio 3.18 to dropdown

almahmoud:
Notable changes:
- R 4.3.2
- Bioc 3.18
- Pinning torch to latest version compatible with current CUDA driver in the container

### DIFF
--- a/config/community_images.json
+++ b/config/community_images.json
@@ -10,21 +10,21 @@
 },
 {
     "id": "RStudio",
-    "label": "RStudio (R 4.3.1, Bioconductor 3.17, Python 3.10.12)",
-    "version": "3.17.1",
-    "updated": "2023-09-13",
-    "packages": "https://raw.githubusercontent.com/anvilproject/anvil-docker/main/anvil-rstudio-bioconductor/info/anvil-rstudio-bioconductor-3.17.1-versions.json",
-    "image": "us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor:3.17.1",
+    "label": "RStudio (R 4.3.2, Bioconductor 3.18, Python 3.10.12)",
+    "version": "3.18.0",
+    "updated": "2023-11-07",
+    "packages": "https://raw.githubusercontent.com/anvilproject/anvil-docker/main/anvil-rstudio-bioconductor/info/anvil-rstudio-bioconductor-3.18.0-versions.json",
+    "image": "us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor:3.18.0",
     "requiresSpark": false,
     "isRStudio": true
 },
 {
     "id": "LegacyRStudio",
-    "label": "Legacy RStudio (R 4.2.3, Bioconductor 3.16, Python 3.10.6)",
-    "version": "3.16.1",
-    "updated": "2023-05-12",
-    "packages": "https://raw.githubusercontent.com/anvilproject/anvil-docker/main/anvil-rstudio-bioconductor/info/anvil-rstudio-bioconductor-3.16.1-versions.json",
-    "image": "us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor:3.16.1",
+    "label": "RStudio (R 4.3.1, Bioconductor 3.17, Python 3.10.12)",
+    "version": "3.17.1",
+    "updated": "2023-09-13",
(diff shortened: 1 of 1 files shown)


almahmoud:
I just noticed there seems to be an issue with versioning by the ammonite script that auto-generates the markdown and version changes. It seems last time the version was changed to 1.1.3 instead of 2.2.3, and now bumped to 1.1.4. The manually added env variable was also using the wrong version. The pushed containers have the right versions though, so it's only the markdown that is wrong afaik. The last commit manually fixes it, but the ammonite script might need changing to not introduce this bug again in the future.

terra-jupyter-gatk/Dockerfile
```diff
@@ -80,6 +80,7 @@ COPY cnn-models.patch /etc/gatk-$GATK_VERSION/cnn-models.patch
 RUN patch -u /opt/conda/lib/python3.10/site-packages/vqsr_cnn/vqsr_cnn/models.py -i /etc/gatk-$GATK_VERSION/cnn-models.patch
 
 # Older version of nbconvert fail to convert notebooks to html
+RUN pip3 install --upgrade --force-reinstall bleach==6.0.0
```
lucymcnatt, line 83: @almahmoud when we try to publish this image the GATK image is failing with `No such file or directory: '/opt/conda/lib/python3.10/site-packages/bleach-6.0.0.dist-info/METADATA'`. Is this addition necessary?

almahmoud: This was added by @LizBaldo, not needed for me

LizBaldo: This error appeared before I added that line @lucymcnatt, the PR as is should not be merged as the GATK image cannot be built. This was my attempts at doing a clean reinstall of the offending package

LizBaldo: This is the offending line in the dockerfile that originally installs a corrupted version of the bleach package https://github.com/DataBiosphere/terra-docker/blob/c3fe12bb991d1cc8abfdf0ce2c057626d0143016/terra-jupyter-gatk/Dockerfile#L31

I could not locate the requirements.txt file, but my guess is that versions are not pinned in there, which is causing the issue :(

lucymcnatt: thanks Liz! yeah there's a comment above that saying that it does that install to
> Grab the requirements from the python base image and reinstall them. 
> By comparison, simply copying the python sites-packages can result in issues when uninstalling packages: https://github.com/conda/conda/issues/10357

So that's [this file](https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-python/requirements.txt) which doesn't pin most things....

lucymcnatt: I can pin the version there and try again

almahmoud:
I've also noticed that this previous PR was never merged https://github.com/DataBiosphere/terra-docker/pull/472, so I incorporated the added sysdeps into this one

almahmoud:
Seems like the Bioconductor test was passing until https://github.com/DataBiosphere/terra-docker/pull/472/commits/bc0b2adcd14e8e5d1dc551274e53a8cbbedf5f6b (https://github.com/DataBiosphere/terra-docker/actions/runs/7024360682/job/19112919527) then started failing after https://github.com/DataBiosphere/terra-docker/pull/472/commits/9f02050b0fd5e331b2da9accc9e8613ec547afca, nothing has changed on our side, but it seems like different packages are being installed in the notebook test?

LizBaldo:
> Seems like the Bioconductor test was passing until [bc0b2ad](https://github.com/DataBiosphere/terra-docker/commit/bc0b2adcd14e8e5d1dc551274e53a8cbbedf5f6b) (https://github.com/DataBiosphere/terra-docker/actions/runs/7024360682/job/19112919527) then started failing after [9f02050](https://github.com/DataBiosphere/terra-docker/commit/9f02050b0fd5e331b2da9accc9e8613ec547afca), nothing has changed on our side, but it seems like different packages are being installed in the notebook test?

Yeah, I feel like I am playing whack a mole here. Now the aou and gatk tests are passing but not the bioconductor or python ones anymore, I need to investigate further

LizBaldo:
@almahmoud since there was absolutely no change to the bioconductor image, I suspect something is going on in the GHA with cached packages so I don't trust the tests. I am going to merge this and test manually as part of the leonardo PR.

PR #476, opened 2024-01-12
Update terra-jupyter-aou to Hail 0.2.128

danking:


terra-jupyter-aou/Dockerfile
```diff
@@ -95,7 +95,7 @@ RUN mkdir -p /tmp/plink && \
 
 RUN mkdir -p /tmp/plink2 && \
   cd /tmp/plink2 && \
-  curl -fsSL -o plink2.zip "https://s3.amazonaws.com/plink2-assets/alpha2/plink2_linux_x86_64.zip" && \
+  curl -fsSL -o plink2.zip "https://s3.amazonaws.com/plink2-assets/plink2_linux_x86_64_latest.zip" && \
```
LizBaldo, line 98: @Qi77Qi or @yonghaoy the previous path for plink2 was removed, this is the closest that I could find, but could you double check?

LizBaldo (approved):

PR #483, opened 2024-04-11
[IA-4839] Terra on Azure (ToA) base jupyter docker image

LizBaldo:
JIRA Ticket: https://broadworkbench.atlassian.net/browse/IA-4839

Our current offering of Jupyter Notebook on Terra on Azure is currently done through the Azure DSVM (Data Science VM). This PR is proposing a base docker image (similar to the one provided on Google) as an alternative to the DSVM to provide the following improvements:

- Make the image GPU-enabled, while staying as lean as possible (it is currently at ~3GB uncompressed) so it can be pulled and deployed without needing any form of caching
- Isolate the user python environment from the jupyter server python environment
- Package the Jupyter Terra offering into an isolated container that can be isolated from the other containers (Welder, Proxy, etc.) so we can safely run it as root and let user perform any operation they might need


.github/workflows/test-terra-base-jupyter.yml
```diff
@@ -0,0 +1,89 @@
...
+        chmod a+r "${{ steps.auth.outputs.credentials_file_path }}"
+        docker run \
+          --env GOOGLE_PROJECT \
+          --volume "${{ steps.auth.outputs.credentials_file_path }}":/tmp/credentials.json:ro \
+          --env GOOGLE_APPLICATION_CREDENTIALS="/tmp/credentials.json" \
+          --volume $GITHUB_WORKSPACE/terra-base-jupyter/tests:/tests \
+          --workdir=/tests \
+          --entrypoint="" \
+          terra-base-jupyter:smoke-test \
+          /bin/sh -c "pip3 install pytest; pytest"
```
cpate4, line 89: would it be possible to either drop this into a `scripts/run tests` (or something like this) and/or in combination use `docker compose` to set this up for the next person working on this?

terra-base-jupyter/custom/test/test-requirements.txt
```diff
@@ -0,0 +1,3 @@
+jupyter==1.0.0
+nose==1.3.7
+requests-mock==1.6.0
```
cpate4, line 3: ... how does this relate to what is in `pyproject.toml`?

terra-base-jupyter/custom/README.md
```diff
@@ -0,0 +1,6 @@
+#Before you merge, lint the files with eslint.
+
+#  `brew install eslint`
+#  `brew install npm`
+#  `npm i`
+#  `eslint --fix *.js`
```
cpate4, line 6: does it make sense to either:
- tie this in to a pre-commit local hook?, or
- make the build (GHA) fail if these are not run?

terra-base-jupyter/README.md
```diff
@@ -0,0 +1,91 @@
+# terra-base-jupyter image
+
+This repo contains the terra-base-jupyter image that is compatible with the
+jupyter notebook service in [Terra]("https://app.terra.bio/") called Leonardo.
+The goal is to provide a lean, GPU-enabled base image that can be use for both Terra-suported, and custom image builds.
```
cpate4, line 5:
```suggestion
The goal is to provide a lean GPU-enabled base image that can be use for both Terra-suported and custom image builds.
```

terra-base-jupyter/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+The goal is to provide the user with a base conda environment where they can install any
+package they need, while keeping the jupyter server itself in an isolated virtual environment
+that DSP devs control using poetry as the dependency manager
+
+## Image contents
+
+`terra-base-jupyter` extends an cuda 12.2.0 and Ubuntu 22.04 base image with the minimum 
+requirements necessary to set up Jupyter and provide compatibility with Leonardo and GPU machines.
```
cpate4, line 14: ... curious why we are not starting with the jupyter containers? (e.g. [minimal-notebooks](https://hub.docker.com/r/jupyter/minimal-notebook) or [jupyter-docker-stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html))

just curious - thanks!

terra-base-jupyter/README.md
```diff
@@ -0,0 +1,91 @@
...
+which allows for cross-platforms, deterministic python environments builds.
+
+You can install poetry via brew or curl commands (NEVER USE PIP!):
+
+```bash
+# Via brew
+$> brew install poetry
+# Via curl
+$> curl -sSL https://install.python-poetry.org | python3 -
+```
```
cpate4, line 48: ... it may have changed since this writing, but `poetry`'s recommended install instructions are to use `pipx` now

lucymcnatt (approved):
Generally looks sound and well-commented! Just left a few questions

.github/workflows/test-terra-base-jupyter.yml
LizBaldo: Yep, I would love to brainstorm that with you because that might slightly fall outside of this PR scope. But the question here is how can we easily test these docker images? I am not convinced that the current GHAs that we have is the way to go

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Node >18 (needed for jupyterlab)
+RUN apt-get update && apt-get install -yq --no-install-recommends 
+RUN mkdir -p /etc/apt/keyrings
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg
+
+RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesource.com/node_$NODE_MAJOR.x nodistro main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN dpkg --remove --force-remove-reinstreq libnode-dev
```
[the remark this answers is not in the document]
LizBaldo: Ahh thanks for the nudge, when I did the POC last year, the apt-install of libnode-dev was kinda corrupted and I had to force reinstall it. But that is not pretty at all. I can do another pass at removing it and fix it properly :) I was definitely focusing on the logic below and did not pay enough attention to these installs

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+ENV CONDA_ENV_NAME base-python3.10
+ENV CONDA_ENV_HOME $USER_HOME/.envs/$CONDA_ENV_NAME
+RUN curl -so $HOME/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-py310_23.5.1-0-Linux-x86_64.sh \
+    && chmod +x $HOME/miniconda.sh \
+    && $HOME/miniconda.sh -b -p $CONDA_ENV_HOME \
+    && rm $HOME/miniconda.sh
+ENV PATH "${PATH}:${CONDA_ENV_HOME}/bin"
+
+ENV BASE_PYTHON_PATH $CONDA_ENV_HOME/bin/python
+ENV PYTHONDONTWRITEBYTECODE=true
```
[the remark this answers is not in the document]
LizBaldo: It should in theory let conda know to NOT keep the python byte code around. I will find the doc and link it

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+###################################################
+## The user should have full access to the conda base environment, and can use it directly, or
+## create new conda environments on top of it. The important part is that jupyter IS NOT installed
+## in the base environment to provide isolation between the user environment, and the jupyter server
+## to avoid cross-contamination
+COPY conda-environment.yml .
+RUN conda env update --prefix $CONDA_ENV_HOME --file conda-environment.yml --prune\
+    # Remove packages tarballs and python bytecode files from the image
+    && conda clean -afy \
+    && find $CONDA_ENV_HOME -follow -type f -name '*.pyc' -delete \
```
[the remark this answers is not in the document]
LizBaldo: Good point, likely not, let me try to remove it and see :)

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+# Append POETRY_HOME to PATH
+ENV PATH "${PATH}:${POETRY_HOME}/bin"
+COPY poetry.lock .
+COPY pyproject.toml .
+
+ENV JUPYTER_VIRTUAL_ENV=/usr/jupytervenv
+# Add jupyter virtual environmemt to PATH
+ENV PATH="$JUPYTER_VIRTUAL_ENV/bin:$PATH"
+
+# Install Poetry, set up the virtual environment for jupyter to run and then cleanup / uninstall poetry
```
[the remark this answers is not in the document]
LizBaldo: My thinking is that poetry is the dependency management that we use as DSP devs to set up python and jupyter, but we want to encourage users to use Conda for their own virtual environment. There is nothing preventing a power user from reinstalling poetry later if they want to, but for now I wanted to stay lean and shed a few mb off the container

terra-base-jupyter/custom/README.md
LizBaldo: Likely yes. These are very outdated scripts that I left in there for now, but i am not sure if we even need it. They might be gone in a sprint or two once I am done setting this up on a BEE 🤞

terra-base-jupyter/custom/test/test-requirements.txt
LizBaldo: It does not, for now I am still debating wether or not we need this (this is the leanest possible amount of legacy code that I am moving over from our current terra-docker process just in case).
If we end up using it, then I will streamline the testing and add this to the dev section of our `pyproject.toml`

LizBaldo: Actually, I already moved it 😅 so might as well delete this file altogether, thanks!

terra-base-jupyter/Dockerfile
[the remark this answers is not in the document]
LizBaldo: Yeah, it is due to some conflict between node and libnode-dev, if i cannot fix it, I will at least better document it

```
#11 26.10 Unpacking nodejs (20.12.2-1nodesource1) ...
#11 31.74 dpkg: error processing archive /var/cache/apt/archives/nodejs_20.12.2-1nodesource1_amd64.deb (--unpack):
#11 31.74  trying to overwrite '/usr/include/node/common.gypi', which is also in package libnode-dev 12.22.9~dfsg-1ubuntu3.5
```

terra-base-jupyter/README.md
LizBaldo: Actually it looks like the official installer is still curl? https://python-poetry.org/docs/#installing-with-the-official-installer

terra-base-jupyter/Dockerfile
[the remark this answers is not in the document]
lucymcnatt: fair enough! Yeah in my mind poetry is so paired with python use, but it makes sense that our users are used to Conda

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+ENV PYTHONDONTWRITEBYTECODE=true
+
+###################################################
+# Set up the user to use the conda base environment
+###################################################
+## The user should have full access to the conda base environment, and can use it directly, or
+## create new conda environments on top of it. The important part is that jupyter IS NOT installed
+## in the base environment to provide isolation between the user environment, and the jupyter server
+## to avoid cross-contamination
+COPY conda-environment.yml .
```
SHuang-Broad, line 123: If I understand correctly the goal here, maybe using conda-pack can help further reduce the image size.
When we build dockers that uses conda to manage envs, we use conda-pack combined with multi-stage builds to reduce image sizes.
Example here:
https://github.com/broadinstitute/long-read-pipelines/blob/4b50b3857d33fd195461e5eb5c8a83d7fe6dda27/docker/lr-papermill-base/Dockerfile#L8

SHuang-Broad (commented):
Thank you so much for the heroic work in putting all these together @LizBaldo !

I am no expert in the Terra IA system, so only focused on the dockerfile itself.

The biggest concern I have, which I don't have a solution for, is python 3.10 is hardcoded in many places, which might make maintenance a little difficult down the road.

terra-base-jupyter/Dockerfile
SHuang-Broad: To add to this, it'd be good to have an understanding how large each layer is, before optimizing the sizes further.

terra-base-jupyter/Dockerfile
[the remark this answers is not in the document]
LizBaldo: Sorry just realized that I replied to the wrong comment 😵‍💫 But I am cleaning up and removing a lot of these and not running into the error anymore

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+RUN apt-get update && apt-get install -yq --no-install-recommends \
+    sudo \
+    ca-certificates \ 
+    curl \
+    jq \
+    # gnupg requirement
+    gnupg \
+    dirmngr \
+    # useful utilities for debugging within the docker
+    nano \
```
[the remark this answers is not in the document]
LizBaldo: Let me try to add them. I am debating between having a super lean base image without much, and letting you add these to your custom docker image instead 🤔

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+
+###################################################
+# Set up the user to use the conda base environment
+###################################################
+## The user should have full access to the conda base environment, and can use it directly, or
+## create new conda environments on top of it. The important part is that jupyter IS NOT installed
+## in the base environment to provide isolation between the user environment, and the jupyter server
+## to avoid cross-contamination
+COPY conda-environment.yml .
+RUN conda env update --prefix $CONDA_ENV_HOME --file conda-environment.yml --prune\
```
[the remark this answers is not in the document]
LizBaldo: I am not sure, let me try it out, thanks!

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,199 @@
...
+# # Terra-specific Jupyter Utilities
+# ##################################
+# Ensure this matches c.ServerApp.port in 'jupyter_server_config.py'
+ENV JUPYTER_PORT 8888
+EXPOSE $JUPYTER_PORT
+ENV JUPYTER_HOME  $JUPYTER_VIRTUAL_ENV/etc/jupyter
+
+# Install the custom extensions to enable welder for file syncing
+COPY custom $JUPYTER_HOME/custom
+COPY custom/jupyter_delocalize.py $JUPYTER_VIRTUAL_ENV/lib/python3.10/site-packages
```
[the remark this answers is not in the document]
LizBaldo: Indeed, thanks for the nudge, I will clean this up and only have the python version definition in one place at the top

terra-base-jupyter/Dockerfile
LizBaldo: That is a good call out. I will add a detail of how big each layer is in the Readme. Regarding conda-pack, I definitely get why you are using it, but since the plan is to have the majority of Terra user use this base image, I was thinking about setting up the base conda environment directly.
The use case of building a custom image on top of it is a bit of an edge case, and so the base image might not be perfectly curated for your need. Would this be a problem?

terra-base-jupyter/Dockerfile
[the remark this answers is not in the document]
SHuang-Broad: totally understand. the assumption is obviously that `less` brings negligible size bloat.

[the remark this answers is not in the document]
LizBaldo: Done :)

terra-base-jupyter/Dockerfile
SHuang-Broad: No, this would not be a problem, just an optimization (which could be evil as it might be pre-mature). OTH, our experience with `conda-pack` is that it reduces the image size further. 
This is where we borrowed the lesson. https://pythonspeed.com/articles/conda-docker-image-size/

I can give it a try once the other components are steady, actually since I'm the one mostly interested in it.

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,198 @@
...
+
+ENV JUPYTER_HOME /usr/jupytervenv
+# Add jupyter virtual environmemt to PATH,
+# but make sure to add it at the end so that the 
+# Conda base python takes precedence 
+# (aka the ! operator in iPython shells should NOT access the jupyter virtualenvironment)
+ENV PATH "${PATH}:${JUPYTER_HOME}/bin"
+
+# Install Poetry, set up the virtual environment for jupyter to run and then cleanup / uninstall poetry
+RUN curl -sSL https://install.python-poetry.org | POETRY_HOME=$POETRY_HOME $BASE_PYTHON_PATH \
```
LizBaldo, line 159: One thing that I was thinking about is maybe not using the user python to install the jupyter server, but instead use the system one to provide another layer of isolation

terra-base-jupyter/Dockerfile
```diff
@@ -0,0 +1,198 @@
...
+#######################
+ENV DEBIAN_FRONTEND noninteractive
+ENV LC_ALL en_US.UTF-8
+
+# We need node >18 for jupyter to work
+ENV NODE_MAJOR 20
+
+# Set the python version and corresponding conda installer
+ENV PYTHON_VERSION 3.10
+ENV CONDA_INSTALLER https://repo.anaconda.com/miniconda/Miniconda3-py310_23.5.1-0-Linux-x86_64.sh
```
LizBaldo, line 19: Look into this as well: https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community

cindy-broadinstitute (approved):

PR #497, opened 2024-12-04
Release Bioc 3.20 RStudio and Jupyter updates

almahmoud:


LizBaldo (approved):

config/community_images.json
```diff
@@ -9,22 +9,22 @@
...
 {
     "id": "LegacyRStudio",
-    "label": "RStudio (R 4.3.3, Bioconductor 3.18, Python 3.10.12)",
-    "version": "3.18.1",
-    "updated": "2024-05-13",
-    "packages": "https://raw.githubusercontent.com/anvilproject/anvil-docker/main/anvil-rstudio-bioconductor/info/anvil-rstudio-bioconductor-3.18.1-versions.json",
-    "image": "us.gcr.io/broad-dsp-gcr-public/anvil-rstudio-bioconductor:3.18.1",
+    "label": "RStudio (R 4.4.1, Bioconductor 3.19, Python 3.10.12)",
+    "version": "3.19.1",
+    "updated": "2024-06-27",
```
lucymcnatt, line 25: nit, shouldn't this be 11-20 as well?

LizBaldo: That is the legacy image though, not sure it needs updating

almahmoud: I went by the date when the image was built. I don't think this is propagated to the UI though anyway, so it doesn't really matter

aednichols (approved):

PR #501, opened 2025-01-14
[AN-332] Specify Legacy in Legacy Rstudio label

LizBaldo:
JIRA ticket: https://broadworkbench.atlassian.net/browse/AN-332

When working on the [Terra UI PR](https://github.com/DataBiosphere/terra-ui/pull/5217), I realized that the labeling of the legacy image was not obvious, and so making it explicit here.

### DIFF
--- a/config/community_images.json
+++ b/config/community_images.json
@@ -20,7 +20,7 @@
 },
 {
     "id": "LegacyRStudio",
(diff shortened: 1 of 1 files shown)


almahmoud (approved):
Thank you for catching that!!

jgainerdewar (approved):

PR #502, opened 2025-02-24
RW-12661 update qiime

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-12661

### DIFF
--- a/terra-jupyter-aou/Dockerfile
+++ b/terra-jupyter-aou/Dockerfile
@@ -186,6 +186,12 @@ RUN R -e 'BiocManager::install(c("GENESIS"))'
 
 RUN R -e 'BiocManager::install(c("SAIGEgds","SNPRelate"))'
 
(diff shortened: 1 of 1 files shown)


terra-rstudio-aou/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.1.2 - 2025-03-07
```
LizBaldo, line 3: Should this changelog and version control be for terra-jupyter-aou instead? I am alsways a bit confused about version control since there is a single image for both jupyter and r.

Would you be able to update the conf.json file accordingly?

LizBaldo (commented):
You can use the GHA to build and publish the aou image you want to test to GCR, and then open a PR in Leo to test on a BEE. 
That way you can merge this PR once your testing is done, let me know if you have any questions

Qi77Qi:
ah cool...forgot it's a GHA for building the image now...will try that

PR #521, opened 2026-07-02
CTM-575 Preparations to update `terra-base`

aednichols:
- Add new Action to build & test `terra-base`, previously did not seem to be tested
- Hop-on: fix bash list syntax that could cause a bug

### DIFF
--- a/terra-base/tests/smoke_test.py
+++ b/terra-base/tests/smoke_test.py
@@ -1,0 +1,87 @@
+"""Smoke tests for the terra-base image.
+
+Verifies that both the user-facing conda kernel env and the Jupyter server venv run the
+image's target Python (the PYTHON_VERSION set in the Dockerfile), and that the Jupyter
+stack and the Terra (Leonardo) notebook extensions import on that Python.
+
+Has no third-party dependencies, so CI can run it without installing anything:
+
+    docker run --rm --entrypoint="" terra-base:smoke-test \
+        /etc/jupyter/bin/python /tests/smoke_test.py
+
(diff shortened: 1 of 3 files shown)


terra-base/tests/smoke_test.py
```diff
@@ -0,0 +1,87 @@
...
+Run it with the Jupyter *server* venv python (it shells out to check the kernel env too).
+The functions are also pytest-discoverable if pytest is available in that interpreter.
+"""
+import os
+import subprocess
+import sys
+
+# The Jupyter server runs in this venv; user notebook code runs in the conda kernel env.
+SERVER_PY = "/etc/jupyter/bin/python"
+KERNEL_PY = "/home/jupyter/.envs/python3/bin/python"
```
LizBaldo, line 21: Is this really the location? I thought it was under a python 3.10 specific name

terra-base/tests/smoke_test.py
```diff
@@ -0,0 +1,87 @@
...
+    got = _minor(SERVER_PY)
+    if target:
+        assert got == target, f"jupyter server venv is Python {got}, expected {target}"
+
+
+def test_server_and_kernel_agree():
+    assert _minor(SERVER_PY) == _minor(KERNEL_PY), "server venv and kernel env Python differ"
+
+
+def test_jupyter_stack_imports():
```
LizBaldo, line 56: For notebooks and labs I would test that you are getting the expected version. Terra is currently not setup for notebook 7 for instance

